PR #111, opened 2014-07-08
Omit non-printable characters with widechar curses

horazont:
This patch replaces nonprintable characters with '?' (questionmark), just like ps and htop without widechar curses handle this.

This prevents possible spoofing attacks using control characters to misplace output (in my tests, using a nonprintable character made the cursor move one character to the left, thus I was able to override other colmuns contents using my process’ name).


hishamhm:
Thank you and sorry about the delay!

PR #113, opened 2014-07-13
minor manpage mistake

zaz:
> Only root can lower the [nice] value.

This confused me when I first read it. By default, any user can lower the nice value, but not below 0.


hishamhm:
Thank you!

PR #122, opened 2014-08-14
Add a 'kill process group' feature bound to the 'g' key 

Arkanosis:
Hello,

The first changeset adds a 'kill process group' feature bound to the 'g' key. It works exactly as the 'kill' feature (F9/k), except that it kills all processes sharing the same PGRP instead of a single process.

The second changeset is purely cosmetic and removes trailing whitespaces (mainly to avoid useless diffs in changesets that actually add some value).

The third changeset fixes a few warnings with GCC 4.8.2 by correctly handling return values (when applicable, meaning that there are still two warnings when handling the return values really makes no sense).

Best regards,


Arkanosis:
Hello,

I've ported the patch to the current version of htop.

It's now a single commit which only adds the “kill group” feature bound to the “g” key. It works exactly as the “kill” feature (F9/k), except that it kills all processes sharing the same PGRP instead of a single process.

Best regards,

hishamhm:
Hi!

The PR looks weird: it shows 139 changed files, and I can't read the diff here on Github ("Files changed" tab says it is truncated). Maybe you should close this and open a new one with the patch alone so I could review it? Thank you!

Arkanosis:
Oh yes, it seems Github has kept the old hishamhm/htop/master as common ancestor :/

Arkanosis:
I've created #368. Thanks!

PR #134, opened 2014-10-08
Interpret command line PIDs as PIDs, not TIDs

jeffgarrett:
The `-p` command line option takes TIDs, but arguably PIDs would be more useful. Here are some reasons why:

 * Personally, I often want to look at a particular process, and much less often a particular thread. 
 * It can be confusing to specify a PID and see only the main thread, especially if the user is unaware there may exist more threads.
 * Threads may come and go, so a list taken by the user before starting htop may not be accurate later.


hishamhm:
Makes sense, I'll give this a try (users can hide threads pressing <kbd>H</kbd>). If this causes confusion, I'll revert it in a future release.

PR #140, opened 2014-10-18
Added new color for 'Z' state.

Kayvlim:
Just like pull request #137, I now added the same color for zombie processes.
I'm not sure whether I should have created a new color `PROCESS_Z_STATE` or if both would share the same format, and opted to make both states highlight the same way.

It currently looks like this:

2014-10-18-024708_138x152_scrot

(I was actually making a patch to allow zombie processes to appear even when kernel threads were hidden, but then I found out Debian Squeeze's htop is from 2008 and that bug has since been fixed.)


Kayvlim:
It's been almost a year! Do you consider this, or a similar option, worth it?

hishamhm:
Oh my, this totally went past me twice!! So sorry about that! 

On reviewing this: I don't think `PROCESS_BAD_STATE` is a good name, since disk activity is not a bad state per se. How about `PROCESS_ALERT_STATE`? If you can get the name revised and the PR in a mergeable state again, I'll merge this right away!

And you're still entitled to complain to me at FOSDEM! :)

Kayvlim:
`PROCESS_ALERT_STATE` sounds good to me too.
I can't resolve the conflicts right now, but I'll send a new PR as soon as I'm able to.

Chances are, it will happen during FOSDEM, and I'll make sure I'll let you know right away! :smiley:

PR #143, opened 2014-10-26
Units and cleanup

eworm-de:
* show units with "i" to indicate binary base
* use define for const values

### DIFF
--- a/MemoryMeter.c
+++ b/MemoryMeter.c
@@ -26,7 +26,13 @@ int MemoryMeter_attributes[] = {
 
 static void MemoryMeter_setValues(Meter* this, char* buffer, int size) {
    Platform_setMemoryValues(this);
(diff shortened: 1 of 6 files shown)


Earnestly:
Hm, why lowercase `k` in `kiB`?
Nice changes otherwise.

eworm-de:
Good question... You are right, for IEC units the `K` in `KiB` is written as a capital letter...

Earnestly:
If you don't mind, I'd be grateful if you could change `k` to `K`.  I could
then make use of this PR for my own custom set of patches against htop
(appending .patch to the PR url) as the developer seems to have no interest
doing this.

eworm-de:
Reworked the code and sent pull request #208.

@Earnestly, happy with that?

Earnestly:
@eworm-de No, not really as it uses incorrect prefixes now.  Thanks anyway.

PR #144, opened 2014-11-04
Add a theme for people using solarized colorscheme

kyzh:
Solarized is a colorscheme for terminals and applications alike.
 
Users of solarised in htop cannot see the "(Black,Black)" in any theme.
This is an attempt to provides those users a way to see all information provided by the default htop theme.

This is a screenshot of my proposed theme and the default one.
screenshot

To achieve this currently, users need to get htop source, modify and recompile htop. 
Alas recompiling (htop) is not a possibility in some environments.

I acknowlege that solarized is not the first nor the last colorscheme available and one might not want to clutter htop with many schemes.
Being able to build a custom color scheme at runtime from htop itself or a .htop would probably be a more elegant way to solve the issue.
Unfortunatly I don't have the technical know-how to provide such a contribution.

Credit:
I used alexeiz's gist for inspiration https://gist.github.com/alexeiz/4657334.

### DIFF
--- a/CRT.c
+++ b/CRT.c
@@ -28,6 +28,7 @@ in the source distribution for its full text.
 #define COLORSCHEME_BLACKONWHITE2 3
 #define COLORSCHEME_MIDNIGHT 4
 #define COLORSCHEME_BLACKNIGHT 5
+#define COLORSCHEME_SOLARIZED 6
 
 #define Black COLOR_BLACK
 #define Red COLOR_RED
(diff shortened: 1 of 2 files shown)


zoresvit:
Is it possible to merge this? Or make any other change so that Solarized colorscheme shows correct?

`htop` is awesome, so is Solarized theme. I'd hate to have compile `htop` with custom patch every time.

Thanks for consideration!

hishamhm:
This is really a bug in Solarized (many apps are broken in this way by it), but in any case I added a "Broken Gray" theme for situations like this. Thank you all for your feedback.

zoresvit:
Oh, I see. Apparently I just haven't happened to use those apps that are broken too.

Anyway, thanks for clarifying.

PR #146, opened 2014-11-15
Automatically follow on process accross tree view / sort order changes

### LINKED ISSUE #51: Processes are not followed correctly
I'm using the current git of htop and on pressing F5 to enable the tree view the focus stays always on the same line (which is mostly not the same process) instead of following the correct process.

cben:
The first commit closes #51.  I think that's a clear win - it's useful to follow a busy process to/from tree view to find its children/parent, and I can't think of any use for staying at the same screen position.
It also keeps following after the switch.  Let me know if you prefer to restore follow=false afterwards, but I think that'd be worse - e.g. going from tree view to CPU sort (which jumps around a lot), you'd follow the current process only to lose it a second later.

The second commit also enables following on any sort order change (if I covered all code paths that do that...).
I'm not sure that's always desired.  The main use case for _not_ following is staying at the top of the screen to look at worse offenders for each resource (which is certainly common use of top!).  But if following is default, one can always go to top of the list with `Home`, whereas if you have to press `F` before switching but forgot, there is no easy way to undo losing the selected process.
The most dubious case is following on inverting sort order (`I`) — the likely reason to press it is wanting to look at opposite processes, not same process on other end of screen... 

Finally, if you like both changes, the `continue` vs `break` structure (the latter implying `follow = false;`) is IMHO getting messy.
It made more sense when most cases that did change it were to false, but now it's more symmetric. 
I'd refactor to explicitly assigning false/true in every case that changes it — let me know what you prefer.

P.S. Thanks for htop, it's the first thing I install on every machine I use!

### DIFF
--- a/htop.1.in
+++ b/htop.1.in
@@ -140,8 +140,8 @@ Sort by time (top compatibility key).
 "Follow" process: if the sort order causes the currently selected process
 to move in the list, make the selection bar follow it. This is useful for
 monitoring a process: this way, you can keep a process always visible on
-screen. When a movement key is used, "follow" loses effect.  When toggling
-tree view, follow is automatically enabled.
+screen. The current process is always followed on any change to sort order,
+or when toggling tree view. When a movement key is used, "follow" loses effect.
 .TP
 .B K
 Hide kernel threads: prevent the threads belonging the kernel to be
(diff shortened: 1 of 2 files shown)


cben:
Any news on the reorganization?
I just compiled from HEAD and see that by default changing sort/tree still doesn't follow.
Would you like an updated PR?

hishamhm:
Hi, yes, an updated PR would be great. The code is really different now!

PR #148, opened 2014-12-03
[PATCH] New setting: "Show program path"

nckx:
Add a setting to hide all but the last component from the programme
path, leaving only the "basename". Makes htop more usable on smaller
screens, or systems with longer than average paths. Off by default.

"Highlight program basename" will still be respected, to further
visually separate process names from their arguments.

Tested on Linux, but doesn't touch anything OS-specific.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -211,6 +211,11 @@ static Htop_Reaction actionToggleUserlandThreads(State* st) {
    return HTOP_RECALCULATE | HTOP_SAVE_SETTINGS;
 }
 
+static Htop_Reaction actionToggleProgramPath(State* st) {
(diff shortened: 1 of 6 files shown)


nckx:
Ping. Is there any interest in something like this?

nckx:
Ping @hishamhm: is there any interest in something like this?

nckx:
Anyway, rebased to `master` FWIW...

grzegorz-dubicki:
+1

hishamhm:
Merged, thank you!

nckx:
Thank you! (Also for the clarification commits.)

Issue #67, opened 2014-04-24
Enhancing the memory meter


Sworddragon:
I'm using the current git of htop and have thought a little about the memory meter:
- The amount of memory is the same as MemTotal from /proc/meminfo. In my case htop is showing me 16048 MiB while 16384 MiB are installed. Currently I'm assuming the missing 336 MiB are reserved by the BIOS + the kernel. Maybe htop is able to determine the physical installed memory and could replace the current value with it while showing the difference as "reserved" or with a similar name.
- The slab cache is counted towards "used". In my opinion the slab cache should get its own entry like "buffers" and "cache" as it is very similar to them (like "buffers" and "cache" it gets dropped if  the correct value is passed to /proc/sys/vm/drop_caches but interestingly it doesn't get much freed if the system runs out of physical memory (but I'm assuming that this is a bug and have opened a bug report for this: https://bugzilla.kernel.org/show_bug.cgi?id=74771 )) and not informative about the real used memory.

MartinSavc:
What was the final decission on this issue?

Depending on the system, slab can grow to considerable size. I have 16Gb of memory, which doesn't get much use lately. When updating my system to Fedora 22, htop started showing 13Gb+ memory used, with system more or less idling. It seems the new kernel has started making more use of the available space, which in meminfo shows up as Slab, most of it as SReclaimable. HTop rather unforunately assigned this to used memory instead of cache (or maybe something more specific).

hishamhm:
Hi, sorry about the long delay in answering this. I am reopening this and I will take a closer look at this issue of memory reporting.

OmegaPhil:
This is not a 'wishlist' item, this is a bug - htop is misreporting a cache (reclaimable part of the kernel slab, SReclaimable) as part of actual used memory - see [the recent change to free for presumably the best example](https://gitlab.com/procps-ng/procps/commit/05d751c4f076a2f0118b914c5e51cfbb4762ad8e).

Sorry I wasn't aware this had been reported before (and even a pull request!), my original comments [here](https://github.com/hishamhm/htop/issues/242#issuecomment-158660553).

OmegaPhil:
Is there anything I can do to get the fix (or similar) into mainline?

IgnorantGuru:
OmegaPhil does meticulous work, just to recommend.  I would like to see this aspect of htop resolved as well - thanks for looking at it.

PR #151, opened 2014-12-19
Show slab usage in memory meter

ypresto:
Slab is a kind of kernel memory which includes reclaimable usage. So it should be shown in different color.

https://github.com/hishamhm/htop/issues/67#issuecomment-49420284

Why I think slab usage should be separated is:

- It is just helpful if kernel, special mem usage is distinguished from user land one.
- Large reclaimable slab usage caused problem in production server monitoring, finally. [See newrelic discussion](https://discuss.newrelic.com/t/memavailable-vs-memfree/12760).
- Latest code of `free` command from procps doesn't includes slab usage in "used" (actually shown as cache == page cache + slab). ([commit](https://gitorious.org/procps/procps/commit/6cb75efef85f735b72e6c96f197f358f511f8ed9))

Many thanks :)


OmegaPhil:
Note that free has [recently been fixed to not include SUnreclaim in the slab cache](https://gitlab.com/procps-ng/procps/commit/05d751c4f076a2f0118b914c5e51cfbb4762ad8e).

hishamhm:
@OmegaPhil, so an ideal PR would be a version of this using SReclaimable instead of Slab?

OmegaPhil:
Basically.

There is one outstanding thing I'm not yet fully aware of - conky is also treating Shmem as allocated memory that isn't counted currently - do you know how this works? I did some digging [on the conky ticket](https://github.com/brndnmtthws/conky/issues/130#issuecomment-158790552) but its an open question for me. free doesn't specially deal with shmem but then it was completely wrong until recently too.

hishamhm:
> do you know how this works?

No, sorry. I haven't kept up with these things lately and each program seems to calculate things differently nowadays, so I'm keen on taking the community's advice.

OmegaPhil:
OK. For reference, at this instant on this machine thats been up ~2d:

```Shmem:            331352 kB
Slab:            3970204 kB
SReclaimable:    3788044 kB
SUnreclaim:       182160 kB```

Not taking into account SReclaimable is a much bigger problem than the ~324MB Shmem here (although if it is wrong, 324MB wrong is still pretty bad) - I'd recommend making the SReclaimable change now, and if I get a response (or find something out) via conky I'll feed back to you later.

hishamhm:
@OmegaPhil, ok! Will do. I posted a comment on shmem in the other issue.

OmegaPhil:
Cheers, sorry I've been stuck on a problem in one of my projects for hours - should be back on this tomorrow.

ypresto:
It seems to be solved in #313

PR #43, opened 2014-04-11
Fix tree view if userland threads are hidden.

ErkiDerLoony:
Hi Hisham,

if userland threads are hidden the tree view of processes contains abundant vertical lines. This should fix it. :)

Best regards

Edgar

PR #162, opened 2015-01-26
Fixed tree drawing for hidden elements

main--:
The tree view has to treat the last element differently ('├' vs '└'). Previously, it would simply pick the last element in the children list, leading to incorrect display if that one happened to be hidden (e.g. because of the "hide userland threads" config option).

### DIFF
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -286,7 +286,14 @@ static void ProcessList_buildTree(ProcessList* this, pid_t pid, int level, int i
          Vector_add(children, process);
       }
    }
+   
    int size = Vector_size(children);
+   
+   int lastShown = -1;
+   for (int i = 0; i < size; i++)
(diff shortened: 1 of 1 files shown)


main--:
I should have taken a look at the open pull requests earlier, as I just noticed that #43 fixes the exact same issue.

>I need to be extra careful when reviewing this one because the memory
management in the juggling between the two vectors is delicate

If you don't want to worry about memory management, you can safely merge this since it doesn't modify the vectors at all.

hishamhm:
Thank you! And sorry for the very late reply. Your solution is indeed simpler, but it makes an extra run through the list... I think the best thing to do would be really to properly review the fix in #43...

hishamhm:
I finally reviewed #43 and merged that one instead. Still, thank you for the contribution!

PR #184, opened 2015-04-20
Add 'Hide program path' option. Change the way basename is calculated.

popizdeh:
Proof of concept. I haven't touched BSD so it won't even compile at this point.

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -87,6 +87,7 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager* 
    Panel_add(super, (Object*) CheckItem_new(strdup("Hide userland threads"), &(settings->hideUserlandThreads), false));
    Panel_add(super, (Object*) CheckItem_new(strdup("Display threads in a different color"), &(settings->highlightThreads), false));
    Panel_add(super, (Object*) CheckItem_new(strdup("Show custom thread names"), &(settings->showThreadNames), false));
+   Panel_add(super, (Object*) CheckItem_new(strdup("Hide program path"), &(settings->hideProgramPath), false));
    Panel_add(super, (Object*) CheckItem_new(strdup("Highlight program \"basename\""), &(settings->highlightBaseName), false));
    Panel_add(super, (Object*) CheckItem_new(strdup("Highlight large numbers in memory counters"), &(settings->highlightMegabytes), false));
    Panel_add(super, (Object*) CheckItem_new(strdup("Leave a margin around header"), &(settings->headerMargin), false));
--- a/Process.c
+++ b/Process.c
@@ -270,24 +270,18 @@ void Process_printTime(RichString* str, unsigned long long totalHundredths) {
 }
 
 static inline void Process_writeCommand(Process* this, int attr, int baseattr, RichString* str) {
+   int basenameLen = this->basenameEnd - this->basenameStart;
    int start = RichString_size(str);
-   RichString_append(str, attr, this->comm);
-   if (this->settings->highlightBaseName) {
-      int finish = RichString_size(str) - 1;
-      if (this->basenameOffset != -1)
-         finish = (start + this->basenameOffset) - 1;
-      int colon = RichString_findChar(str, ':', start);
-      if (colon != -1 && colon < finish) {
-         finish = colon;
-      } else {
-         for (int i = finish - start; i >= 0; i--) {
-            if (this->comm[i] == '/') {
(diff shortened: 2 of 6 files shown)


grzegorz-dubicki:
Can you show the expected results of your PR? Maybe this could help in resolving https://github.com/hishamhm/htop/issues/233.

popizdeh:
It might alleviate the problem if the path is long but I doubt it. It'll show chrome instead of /opt/google/chrome/chrome, and will properly highlight program name.

nckx:
https://github.com/hishamhm/htop/pull/148 :cry:

popizdeh:
I tried your patch but the basename calculation didn't work well if I remember correctly :(

nckx:
Well, I re-used the exisiting basename calculator, so I'd be very interested in where exactly the problem is. Do you happen to have more information, or even a reproducer?

If not, no problem; but I've been using that patch since 2014 (yikes) and haven't noticed problems or got any feedback on that PR. I can't fix what no-one tells me about :confused:

popizdeh:
These are some examples

*kworker/3*:OH

watchdog/*5*

/opt/google/chrome/chrome --type=renderer --enable-deferred-image-decoding
--lang=en-GB
--force-fieldtrials=*AffiliationBasedMatching/Enabled/AudioProcessing48kHzSupport/Default/*AutofillEnabled/Default/CaptivePortalInterstitial/Enabled/*ChildAccountDetection/Disabled/ChromeDashboard/Default/*DomRel-Enable/enable/EnableSessionCrashedBubbleUI/Disabled/*EnhancedBookmarks/Default/*ExtensionContentVerification/Enforce/*ExtensionDeveloperModeWarning/Default/*ExtensionInstallVerification/None/*IconNTP/Default/*NewProfileManagement/Enabled/*NewVideoRendererTrial/Enabled/*OmniboxBundledExperimentV1/Stable_HQPDemotion_HUPDecay_Enabled_V3/*PasswordGeneration/Disabled/PasswordLinkInSettings/Enabled/*PermissionBubbleRollout/Enabled/*PrerenderFromOmnibox/OmniboxPrerenderEnabled/*QUIC/EnabledNoId/*RefreshTokenDeviceId/Enabled/*RememberCertificateErrorDecisions/Default/ReportCertificateErrors/ShowAndPossiblySend/ReportCertificateErrorsOverHttp/UploadReportsOverHttp/SHA1IdentityUIWarning/Enabled/SHA1ToolbarUIJanuary2016/Warning/SHA1ToolbarUIJanuary2017/Error/*SafeBrowsingIncidentReportingService/Default/*SdchPersistence/Default/SessionRestoreBackgroundLoading/Restore/*SettingsEnforcement/no_enforcement/*SyncBackingDatabase32K/Enabled/*UMA-Dynamic-Binary-Uniformity-Trial/default/*UMA-Population-Restrict/normal/*UMA-Uniformity-Trial-1-Percent/group_27/*UMA-Uniformity-Trial-10-Percent/group_09/*UMA-Uniformity-Trial-100-Percent/group_01/*UMA-Uniformity-Trial-20-Percent/group_03/*UMA-Uniformity-Trial-5-Percent/group_12/*UMA-Uniformity-Trial-50-Percent/group_01/*UseDelayAgnosticAEC/DefaultEnabled/*VoiceTrigger/Install/
* --v8-natives-passed-by-fd --v8-snapshot-passed-by-fd
--enable-offline-auto-reload --enable-offline-auto-reload-visible-only
--enable-pinch-virtual-viewport --enable-delegated-renderer
--num-raster-threads=2 --gpu-rasterization-msaa-sample-count=8 --use-ima*

Have a look at my patch, I don't even remember what I did. I just know that
main problem is that basename calculation search for last occurrence of
slash which is too naive. My patch isn't bulletproof either because the
APIs suck, I don't think you can ever be 100% correct if all you have is
the full command. Ideally you'd like the OS to tell you what part of that
path is directory, file and arguments.

hishamhm:
Hi, sorry for being silent for so long. I've been working on a lot of stuff at the same time and haven't been able to give htop priority for an appropriate stretch of time.

As you said, there's no bulletproof way to determine the basename. @popizdeh, if I follow the code correctly your approach fixes the issue of the slash... but breaks names with spaces. I think @nckx's code works with spaces but assumes slashes are path separators, as in the original code.

I think I'll go with @nckx's code which is closer to the current behavior and tweak it a bit to make it jump over the calculations when the flags are off (to keep the "fast path" lean).

hishamhm:
Closing this PR now... in any case, thank you for the feedback!

PR #193, opened 2015-05-13
added missing android defines

sherpya:


hishamhm:
Note that the file you're adding to in this PR actually starts with:

```c
/* Do not edit this file. It was automatically generated. */
```

You should edit Process.c instead. Did you try making a full, clean build?

sherpya:
you are right, I've changed it after it failed to build, it's just a bit uncommon to generate a .h that is not a config

PR #194, opened 2015-05-15
fix compiler warnings

eworm-de:
gcc gives warnings like this:

warning: ignoring return value of ‘fscanf’, declared with attribute
warn_unused_result

Assign value to a variable, cast to (void) to discard it.


hishamhm:
Thank you! It's odd that
```c
(void) fscanf(file, "%32d", &maxPid);
```
wasn't enough to make gcc happy.

PR #203, opened 2015-06-07
Fix several FreeBSD compilation errors

Lance0312:
Several functions and structures had their name changed in previous commits and cause compilation errors.


hishamhm:
Oh, thank you! I haven't been able to work on this a lot lately, but I'd like to get a release out soon, with at least minimal support for FreeBSD to serve as an example port.

PR #205, opened 2015-06-10
Change all displayed memory size specifiers to use GNU Coreutils style size specifiers

EliteTK:
These patches should standardise the memory size specifiers used within htop.
It only addresses visual inconsistencies, however, there are still a lot of instances in the htop source where a mix of JEDEC standard and IEC standard is used.


hishamhm:
Thank you!!

PR #208, opened 2015-06-11
Dynamic unit

eworm-de:


eworm-de:
To give an idea of what it looks like:
pr-208

hishamhm:
This is interesting! The more I look at it, the more I get used to it and like it.
Minor nit, though: Meter_human_unit -> Meter_humanUnit

eworm-de:
Ah, sure. Updated to match function naming convention.

hishamhm:
Merged, thanks!!

PR #209, opened 2015-06-14
make meters in graphical mode more colorful

eworm-de:


eworm-de:
To give an idea of what it looks like:
pr-209

The bits are in place for all themes, but I changed the colors for default one only. Feel free to update all themes to your needs.

LukasLedvina:
Nice thing, I'm using that :). I hope it will be better to have a possibility to change colors when a different things are measured (i.e. CPU & Memory usage for example). I don't know how to provide that in current code.

eworm-de:
The default theme now looks like this:
screenshot_2015-08-20_08-53-18

The font is [Terminus](http://terminus-font.sourceforge.net/), which includes bitmap braille characters since version 4.40. Yeah!

hishamhm:
I'm not merging this because the colors may mislead one into thinking they represent different things, now that the graph presents bars (especially because htop's bar mode does use colors for diferent things. Unfortunately, due to character mode limitations we can't reproduce the meaningful colors in the graph).

eworm-de:
A valid point... Still I think it's useful if you can differentiate between the lines. How about #252?

Explorer09:
Hey guys, would you like to try out my changes? See pull request #327

PR #211, opened 2015-06-15
ignore enter, delete and space on empty panel

### LINKED ISSUE #210: segfault with action on empty panel
Removing all meters from a panel and pressing *Enter*, *Del* or *Space* will crash htop. It tries to act on a non existent meter.

eworm-de:
Should fix #210.


hishamhm:
Good catch, thank you!

PR #213, opened 2015-06-23
LoadAverage-bar scaling

LukasLedvina:
Hi, I'm set the top of the LoadAverage-bar to the #CPUs. Top @ 100 make no sense. And I switched off automatical upscaling.
Future feature request might be upscale the load average as needed and when it decreases downscale it slower to the #CPUs.
Lukas

PS: I fixed one misc.

### DIFF
--- a/LoadAverageMeter.c
+++ b/LoadAverageMeter.c
@@ -25,6 +25,12 @@ static void LoadAverageMeter_setValues(Meter* this, char* buffer, int size) {
    snprintf(buffer, size, "%.2f/%.2f/%.2f", this->values[2], this->values[1], this->values[0]);
 }
 
+static void LoadAverageMeter_init(Meter* this) {
+   int cpus = this->pl->cpuCount;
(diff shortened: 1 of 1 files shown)


LukasLedvina:
I fixed one misc.

hishamhm:
In a heavily loaded system, sometimes the load averages are much higher than the number of CPUs. 100 is too much, indeed, but on the other hand, setting the bar to 1 for a single-processor machine is too low. Maybe something like 5*cpus is reasonable...

LukasLedvina:
Ok, what do you think about this improvement:
  * to make a bar max value dynamically adjustable.
  * start with value 1 per CPU
  * if load is higher than max bar value, set up the max bar value to the load
  * in the opposite case (when max value > #CPU) decrease max bar value by say 1% ?
  * to distinguish between overloaded system and not overloaded system draw bar by '|' bellow #CPU and the rest of the bar by '~', or by an another character.

PR #217, opened 2015-06-24
remove all header file, we generate these on the fly

eworm-de:
No need to have these in git repository, no?


hishamhm:
This is a convenience, to avoid having to have a compatible Python if all you want is to get htop up and running from sources. (I've bumped into that a few times myself when running in random distros.)

eworm-de:
You could remove the files from git repository but include in release tarball.

PR #219, opened 2015-06-25
Add the "Active users" meter.

nojhan:
The list is truncated to the size of the display.

Tested on linux (Ubuntu 14.04).


nojhan:
PR #222 extends this patch with more features.
I let this one opened in case you prefer this simpler implementation.

hishamhm:
This patch has problems. Checking with strstr() doesn't work... This implementation (and PR #222) will miss user 'anna' if user 'adrianna' is already in the list. Also, you're assuming the process list is sorted so that running processes are at the top.

nojhan:
I will fix the implementation for the name search.

Concerning the assumption, aren't the following lines supposed to guarantee that the process list is sorted?
```
pl->settings->sortKey = PERCENT_CPU;
pl->settings->direction = 1;
ProcessList_sort(pl);
```

hishamhm:
Meters shouldn't alter the process list in any way (including re-sorting it).

nojhan:
Willco.

PR #220, opened 2015-06-30
use unicode block elements for graphical meter

eworm-de:
Not sure whether or not this should be merged... The issue with using braille characters is:

* looks ugly if true type font is used, depending on the font it can be even completly useless (with Fira Mono for example)
* breaks completly if no braille characters are available

I wondered whether it makes sense to use [unicode block elements](https://en.wikipedia.org/wiki/Block_Elements) instead:

* no unusable truetype stuff

But the drawback is:

* The resolution is limited

So this is RFC... Waiting for comments. :D

### DIFF
--- a/htop.1.in
+++ b/htop.1.in
@@ -237,7 +237,7 @@ plus twenty. Different for real-time processes.
 .B NICE (NI)
 The nice value of a process, from 19 (low priority) to -20 (high priority). A
 high value means the process is being nice, letting others have a higher
-relative priority. Only root can lower the value.
+relative priority. The usual OS permission restrictions for adjusting priority apply.
 .TP
 .B STARTTIME (START)
 The time the process was started.
@@ -366,7 +366,7 @@ The decision to use this convention was made in order to conserve screen space
 and make memory size representations consistent throughout htop.
 
 .SH "SEE ALSO"
-proc(5), top(1), free(1), ps(1), uptime(1)
+proc(5), top(1), free(1), ps(1), uptime(1), limits.conf(5)
 
 .SH "AUTHORS"
 .LP 


LukasLedvina:
Hi, I'm not testing that, but Meter.c:368 "const double dot = (1.0 / 16);" should be 1.0/8  --- vertical resolution of the graph (and also few other lines, see below). In your solution the bar is increasing from 0% to 12.5%, than is constant up to 25%, increasing to 37.5%, constant to 50%,...

 Meter.c:368 16->8
 Meter.c:375 12->6
 Meter.c:385 4->2

How about introducing variable "pixPerRow" (pixels per console row).
Lukas

eworm-de:
> Hi, I'm not testing that, but Meter.c:368 "const double dot = (1.0 / 16);" should be 1.0/8 --- vertical resolution of the graph (and also few other lines, see below). In your solution the bar is increasing from 0% to 12.5%, than is constant up to 25%, increasing to 37.5%, constant to 50%,...

Ah, of course you are right. This was intended to be a quick'n'dirty base for discussion only, but better do it right. ;)

> How about introducing variable "pixPerRow" (pixels per console row).

Did that in #221.

hishamhm:
@eworm-de the issue with TrueType fonts depends a lot on which fonts you happen to have available, so it's just as true with Braille as it is with block characters. Both sets may or may not be missing.

If you use a font which lacks the necessary characters, it's possible to write a fontconfig configuration file to override that segment of Unicode with characters from another font.

eworm-de:
> the issue with TrueType fonts depends a lot on which fonts you happen to have available, so it's just as true with Braille as it is with block characters. Both sets may or may not be missing.

I've tried Fira Mono in my terminal. That font does have braille characters, but it has dots when set and circles when unset. It's nearly impossible to see a difference on small font size and/or low resolution display. That is why I thought I should try something else that can break less.

However I still prefer braille characters.

> If you use a font which lacks the necessary characters, it's possible to write a fontconfig configuration file to override that segment of Unicode with characters from another font.

Ah, nice! Did not know that. Thanks for the hint!

PR #222, opened 2015-07-03
Color active users according to levels of activity.

nojhan:
This is a follow-up of the text meter that shows the names of users having active processes (Pull request #219, that you can discard if you choose this one).

Adds ACTIVE_USERS_* style fields.


nojhan:
Superseeded.

PR #223, opened 2015-07-07
simplify UTF-8 detection

eworm-de:


eworm-de:
This conflicts with 749198006a18f7812ab79a4ea16753c6a780adee... Any chance to start merging any time soon? Or should I rebase one of both on the other?

hishamhm:
Thank you! I didn't know about langinfo.h. Is CODESET always normalized into "UTF-8" regardless of what users set in their environment variables?

eworm-de:
CODESET always normalized into "UTF-8" if it is UTF-8. AFAIK there are some other charsets that support multi byte character, but that are non-UTF-8. So instead of hard coding UTF-8 characters into the code we would have to handle unicodes (something like 0x28c0 for Braille P78) and convert that via iconv() or similar.

Interested in a patch/pull request?

hishamhm:
I wonder which of these other charsets are in current use in modern systems... I'd rather add the extra complication to the code only if there are actual htop users needing it. So, I'd probably rather get a bug report first about this before I revisit the issue.

PR #232, opened 2015-07-25
check `realloc` for `NULL`

tijko:
I added some similiar checks on `realloc`s return (as in `String.c`)

### DIFF
--- a/Affinity.c
+++ b/Affinity.c
@@ -44,7 +44,9 @@ void Affinity_delete(Affinity* this) {
 void Affinity_add(Affinity* this, int id) {
    if (this->used == this->size) {
       this->size *= 2;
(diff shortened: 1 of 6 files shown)


hishamhm:
Hi,

Thanks for the initiative of helping out! However, checking the result of realloc and then keeping the original memory block if the reallocation failed is just as problematic as not checking the result at all. You're just moving the crash a few lines further in the code.

The decision of not checking mallocs/reallocs was deliberate: if the system is so out of memory that realloc fails, there's nothing useful we can do anymore and it's better to just let htop crash.

tijko:
I realized the crash was delayed but, I wasn't sure If this was intentional.  I thought it might be better to handle this where the resulting `NULL` was originally caught.

PR #234, opened 2015-07-31
Abort after alloc error

tijko:
How do you feel about using assert to abort right after the error?

### DIFF
--- a/Affinity.c
+++ b/Affinity.c
@@ -45,8 +45,8 @@ void Affinity_add(Affinity* this, int id) {
    if (this->used == this->size) {
       this->size *= 2;
       int* tmp_cpus = (int*) realloc(this->cpus, sizeof(int) * this->size);
-      if (tmp_cpus != NULL)
-         this->cpus = tmp_cpus;
+      assert(tmp_cpus != NULL);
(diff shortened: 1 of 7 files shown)


hishamhm:
I know all best practice books say to check the output of malloc and realloc, but really: htop only reads data. Crashing with an assert() right after realloc() is no different from crashing a few lines later when it attempts to use the data — the net result for the user is the same. There is no data corruption, no data loss. If this was, say, a text editor, then we'd have to take care about this, but then simply asserting wouldn't be enough — the correct thing to do would be to implement some sort of recovery from out-of-memory situations, probably using a pre-allocated static buffer. In htop, instead, the is no data to be lost. The user simply restarts htop and carries on.

Also, [assert() is a debugging tool](http://linux.die.net/man/3/assert): if htop is compiled with NDEBUG, the check is disabled. But debug mode is of no use to debug out-of-memory situations, as the memory use characteristics are different from code compiled in production mode.

tijko:
Hi hishamhm, I appreciate the insight.  I had in mind that aborting as soon as possible while maintaining the original data pointer would allow inspection of the stack in the case that any of the data would be of use.

PR #246, opened 2015-08-21
introduce UptimeMeter_display() and display long uptime in red

eworm-de:


eworm-de:
Updated the code to get values just once and store in meter struct.

PR #247, opened 2015-08-25
remove extra space in dual meters without margins

eworm-de:


hishamhm:
Make pad 1 in no-margin mode and I'll merge this :)
We need at least one char cell between columns.

eworm-de:
All meters in bar mode do have at least one trailing space. Isn't that enough?

hishamhm:
Yes, I just tested it here and it looks fine. Thanks!

PR #250, opened 2015-08-25
use ASCII blank even for graphical meter in UTF-8 mode

eworm-de:
If the terminal has no font with braille characters we see... Nothing
useful. Use an ASCII blank at least, so we have an idea about what's
going on.


hishamhm:
Does this produce a visual difference?

eworm-de:
Sure it does produce a visual difference. This is on a system without braille characters:
screenshot_2015-08-27_05-59-03
And with the change applied:
screenshot_2015-08-27_06-00-45

hishamhm:
Oh, right! Thanks for the screenshots! :)

PR #251, opened 2015-08-28
add a project .vimrc file

eworm-de:
We use an unusual indent of three spaces. Let's give vim a hint about
that.


hishamhm:
Thanks @eworm-de and @klaernie! Merged that instead.

PR #254, opened 2015-08-28
kilobytes is the smallest unit and never has precision

eworm-de:


hishamhm:
watch out for the indentation!

eworm-de:
Oops, did I miss that again? You should merge #251. :D

eworm-de:
Updated with fixed indention.

PR #262, opened 2015-09-07
fix calloc() calls

### LINKED ISSUE #261: calloc() problem
It seems like [here](https://github.com/hishamhm/htop/blob/master/Header.c#L92):

```
char* name = calloc(64, sizeof(char*));
```

The `size_t` value should be `sizeof(char)`, not `sizeof(char*)`

Found with the Clang static analyzer.

eworm-de:
* size_t nmemb (number of elements) first, then size_t size
* do not assume char is size 1 but use sizeof()
* allocate for char, not pointer to char (found by Michael McConville,
  fixes #261)

### DIFF
--- a/Header.c
+++ b/Header.c
@@ -89,7 +89,7 @@ void Header_writeBackToSettings(const Header* this) {
       
       for (int i = 0; i < len; i++) {
          Meter* meter = (Meter*) Vector_get(vec, i);
-         char* name = calloc(64, sizeof(char*));
+         char* name = calloc(64, sizeof(char));
          if (meter->param) {
             snprintf(name, 63, "%s(%d)", As_Meter(meter)->name, meter->param);
          } else {
--- a/Panel.c
+++ b/Panel.c
@@ -455,7 +455,7 @@ bool Panel_onKey(Panel* this, int key) {
 HandlerResult Panel_selectByTyping(Panel* this, int ch) {
    int size = Panel_size(this);
    if (!this->eventHandlerState)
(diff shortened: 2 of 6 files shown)


jes:
Just a note to say that sizeof(char) is defined to be 1.

eworm-de:
This is true for linux, but I think there are other architectures that handle this different, no?
Or was it the other way round that sizeof(char) is always 1 but that is not always one byte?

However... Using sizeof() is the save option and it does not bring any extra cost as the compiler optimizes this.

hishamhm:
> This is true for linux, but I think there are other architectures that handle this different, no?

Not in the last 40 years, I think :) Certainly not in any platforms that htop runs or is expected to run in the future ;)

mmcco:
> > This is true for linux, but I think there are other architectures that handle this different, no?

> Not in the last 40 years, I think :) Certainly not in any platforms that htop runs or is expected to run in the future ;)

Regardless, it's *definitely* still preferable to use `sizeof(char)` rather than hardcoding an assumed value. It's more readable, it helps static analysis, and it aids portability when someone wants to do something wacky for fun.

hugomg:
FYI, jes is right. The C standard specifies that sizeof(char) should be 1, even if you are on a wachy machine with 9bit chars or something like that

http://stackoverflow.com/questions/2215445/are-there-machines-where-sizeofchar-1

mmcco:
Yeah, y'all are right. I read the above too quickly and saw it only as a "malloc v. calloc" discussion".

Issue #242, opened 2015-08-18
document how htop calculates used/free memory values


jackyf:
Hi!

A Debian user reported that there is a significant (5-10%) discrepancy between "free + caches/buffers" as reported by 'free' and the numbers reported by htop.

It would be nice for man page to contain a sentence about how memory usage is calculated.

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=795776

hishamhm:
```
Hello,

On 17.08.2015 23:45, OmegaPhil wrote:
> omega@omega1:~/.config/xfce4/panel$ free
>               total        used        free      shared  buff/cache
> available
> Mem:      32998636     9515852      466464      263632    23016320
> 22807536
> Swap:      10485756           0    10485756
> 
> ====================================================================
> 
> free is '/usr/bin/free':
> [...]

I see now, an output format of 'free' changed in sid, I get similar now after I have installed a newer version.

> Atm htop reports 12550/32225MB used.

Yes, I see that numbers don't match exactly on your system.

I forwarded your report upstream.
```

Note that 32998636 / 1024 = 32225: htop reports in megabytes, /usr/bin/free reports in kilobytes.

Apparently /usr/bin/free is summing buffers and cache, so I don't get the separate numbers, but in my version of /usr/bin/free, where I have the "-/+ buffers/cache" column, the first value X that shows on "X/Y MB" in the memory meter is the same as the first value that appears in the  "-/+ buffers/cache" line, divided by 1024 (in other words, used - (buffers + cache) from the first line).

I don't know how this new /usr/bin/free is calculating its numbers, but htop's calculation is straightforward, using numbers from /proc/meminfo. It's basically `(MemTotal - MemFree - (Buffers + Cached)) / 1024`.

jackyf:
Hello,

On 20.08.2015 00:10, Hisham Muhammad wrote:
> I don't know how this new /usr/bin/free is calculating its numbers, but htop's calculation is straightforward, using
> numbers from /proc/meminfo. It's basically |(MemTotal - MemFree - (Buffers + Cached)) / 1024|.

Makes sense. Would be still nice to put the (variation of) last sentence to the man page.


-- 
-- t. Ievgen

hishamhm:
htop calculates various things, not sure if I want to all replicate calculations in the manpage... (e.g. then there are CPU calculations, etc.) I'd rather keep it more abstract.

OmegaPhil:
[...]

PR #263, opened 2015-09-07
add Slab memory to cached

MartinSavc:
I use Fedora 22, kernel 4.1.6.

On my system free and htop report drastically different application memory usage. It comes down to how cached memory is calculated. Free adds Cached and Slab entries from memoryinfo, while htop only reads Cached. 

I've prepared a simple fix.

Regards,
Martin

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -621,6 +621,7 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, const char*
 
 static inline void LinuxProcessList_scanMemoryInfo(ProcessList* this) {
    unsigned long long int swapFree = 0;
+   unsigned long long int slab = 0;
 
    FILE* file = fopen(PROCMEMINFOFILE, "r");
    if (file == NULL) {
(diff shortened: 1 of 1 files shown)


OmegaPhil:
I'm assuming you're still subscribed to this pull request - why was it closed? [It does partly fix a serious problem with htop-reported memory usage](https://github.com/hishamhm/htop/issues/242#issuecomment-158660553), although it needs to use SReclaimable rather than Slab due to the latter's SUnreclaim part.

MartinSavc:
My pull request was a quick fix after a small amount of research for a problem I've noticed on my system. After submitting the request I've took a closer look at past submissions for htop, and others have submitted pull request for complete solutions. Therefore I've removed mine, since it wouldn't make sense to merge it instead of other solutions.
I've noticed my error later on as well, but again I saw no benefit in fixing it as other have proposed a complete solutions before me.

OmegaPhil:
Oh wow, its been sat there for a year?? Sorry I did a search and hit on your pull, thought I got lucky but it seems this problem has been ignored for some time... thanks for responding.

PR #264, opened 2015-09-08
Add the "Active users" meter. 

nojhan:

Tested on linux (Ubuntu 14.04).

Adds ACTIVE_USERS_* style fields.

This is a fixed follow-up of pull requests #219 and #222.


nojhan:
Rebased on #494

PR #267, opened 2015-09-10
Color type

eworm-de:


eworm-de:
Not sure if this is correct at all...
Looking at ncurses header files colors and color pairs are stored in short (16 bit signed). The A_* values have a size of 4, though... But what type are they?

PR #269, opened 2015-09-10
Detect whether to use libtoolize or glibtoolize.

jweyrich:
Answering https://github.com/hishamhm/htop/pull/268#issuecomment-139357377, `glibtoolize` is provided by Homebrew in the `libtool` package.

> $ brew ls libtool | grep /bin/glibtoolize
> /usr/local/Cellar/libtool/2.4.6/bin/glibtoolize

Btw, `libncursesw` doesn't seem to be available in Darwin either (not sure though), so I disabled unicode via `./configure --disable-unicode`. Will need to take a better look later.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,9 +1,23 @@
 #!/bin/sh
 
+# libtoolize verification thanks to libuv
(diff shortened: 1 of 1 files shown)


hishamhm:
I had just written a simpler one!

jweyrich:
Much simpler! :+1:

PR #273, opened 2015-09-14
Mac build fixes

kaefer:
attempt to fix glibtoolize detection, checked both on Kubuntu 15.04 and OS X 10.9.1

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-glibtoolize=$(which glibtoolize)
+glibtoolize=$(which glibtoolize 2> /dev/null)
 if [ ${#glibtoolize} -gt 0 ]
 then libtoolize=glibtoolize
 else libtoolize=libtoolize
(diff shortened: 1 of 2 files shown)


autogen.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 
-if glibtoolize --version &> /dev/null
+glibtoolize=$(which glibtoolize)
```
hishamhm: Does changing this to
```sh
glibtoolize=$(which glibtoolize 2> /dev/null)
```
work as expected? (Correct detection, yet silent)

kaefer: jep, is working, but I don't see the point. 
If `which` doesn't find the binary in PATH, it shows no output

hishamhm: Not the case in my machine (I get an error message in stderr when glibtoolize is not found)

kaefer:
fixed :)

hishamhm:
Thank you!

PR #274, opened 2015-09-17
Cleanup and initial OpenBSD support

mmcco:
The person writing the Darwin code has a pretty strange style, honestly. Not putting spaces after condition keywords, putting 0 or NULL first in comparisons, etc.


darwin/DarwinProcessList.c
```diff
@@ -73,10 +73,8 @@ unsigned ProcessList_allocateCPULoadInfo(processor_cpu_load_info_t *p) {
 void ProcessList_getVMStats(vm_statistics64_t p) {
     mach_msg_type_number_t info_size = HOST_VM_INFO64_COUNT;
 
-    if(0 != host_statistics64(mach_host_self(), HOST_VM_INFO64, (host_info_t)p, &info_size)) {
-       fprintf(stderr, "Unable to retrieve VM statistics\n");
-       exit(9);
-    }
+    if (host_statistics64(mach_host_self(), HOST_VM_INFO64, (host_info_t)p, &info_size) != 0)
+       err(9, "Unable to retrieve VM statistics\n");
```
hishamhm: Please avoid removing braces, even when block contains a single statement

hishamhm:
Thanks for taking a closer look at the Darwin codebase. Is this good to merge?

StringUtils.c
```diff
@@ -55,13 +55,13 @@ inline int String_eq(const char* s1, const char* s2) {
 char** String_split(const char* s, char sep, int* n) {
    *n = 0;
    const int rate = 10;
-   char** out = (char**) malloc(sizeof(char*) * rate);
+   char** out = calloc(rate, sizeof(char**));
```
eworm-de: This one looks wrong. We need sizeof(char*), not sizeof(char**).

StringUtils.c
```diff
@@ -55,13 +55,13 @@ inline int String_eq(const char* s1, const char* s2) {
 char** String_split(const char* s, char sep, int* n) {
    *n = 0;
    const int rate = 10;
-   char** out = (char**) malloc(sizeof(char*) * rate);
```
mmcco: I don't think so... you allocate my_type* for a variable that's my_type**, no?

StringUtils.c
mmcco: This is indeed wrong, I was moving too fast. We need `sizeof(char*)`.

mmcco:
There are too many small mistakes in here to merge. It was a hasty correction of a few issues I came across while porting to OpenBSD. You can consider it a guideline and add them manually, or I can submit a more carefully written patch soon.

mmcco:
I just added my OpenBSD port. I also fixed the `calloc()` mistake I made. I still need to change some of the `err()`s to `errx()`s, but that's minor. A file or two in Darwin may be missing `#include <err.h>`, too. I need to sleep first, though.

hishamhm:
Please make separate PRs, especially for a large feature such as OpenBSD support.
Also, make sure that indentation is consistent with the rest of htop's codebase.
Adding OpenBSD support is a great feature! Thank you :)

And do get some sleep! Health is important :)

PR #281, opened 2015-09-30
Added preliminary attempt at jails support on FreeBSD - JID and JAIL (name) columns

etosan:
also, somewhat more correct kernel "thread" detection.

Seems FreeBSD kernel can spawn both kernel processes (what is what htop currently sees) and kernel threads. For now let's consider kernel processes kernel "threads".

### DIFF
--- a/freebsd/FreeBSDProcess.c
+++ b/freebsd/FreeBSDProcess.c
@@ -12,6 +12,7 @@ in the source distribution for its full text.
 #include "CRT.h"
 
 #include <stdlib.h>
+#include <string.h>
 #include <unistd.h>
 #include <sys/syscall.h>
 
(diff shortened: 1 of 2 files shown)


etosan:
I finally noticed you added support for FreeBSD. So I tried to compile it natively and to my surprise most of it works! Only problem is with python generator - ports default python executable is python2.7. I think some counters are not exactly correct, but majority of it seems to work and with i can actually see jail names and JIDs an I don't need linux emulation procfs mount. OK?

hishamhm:
@etosan Thank you for the PR! The FreeBSD support was a proof-of-concept I wrote to isolate the Linux-specific parts from the rest of the code... I'm happy to see a FreeBSD user give some love to the `freebsd/` codebase!

PR #285, opened 2015-10-06
Unless I move signal definitions into the "generator" comment...

etosan:
htop fails to compile with:

```text
SignalsPanel.c:32:49: error: use of undeclared identifier 'Platform_signals'
      Panel_set(this, i, (Object*) ListItem_new(Platform_signals[i].name, Platform_signals[i].number));
                                                ^
1 error generated.
*** Error code 1
```


hishamhm:
The problem was that the table was marked `static`. Deleting it removes the need of having it in the "generator" section. 

Thanks for catching the "default signal" issue! I fixed it a bit differently.

PR #287, opened 2015-10-06
PPID and JID update fix

etosan:
Sorry for being lame with git.
After some other testing I found error with ppid and jid not matching actual system situation, FreeBSDProcessList.c in 23bf564 contains the fix.


hishamhm:
I didn't run the code, but it looks fine to me, thank you!

PR #291, opened 2015-10-13
adding "SignalsPanel.h" to imports to make it compile again on OS X

kaefer:
Previous commit added only the SignalPanels support without adding SignalItem definition. 
=> compiling fails on OS X.
This is fixed with simply adding the relevant header file to the imports of darwin/Platform.c


hishamhm:
Thank you — I regenerated the headers.

kaefer:
well I'm sorry, but your commit doesn't fix it. I get the same error as before after pulling and trying a clean build.

PR #293, opened 2015-10-17
Fix compilation under OSX/Darwin

Doridian:
by copying Platform.h from linux to add additional externs


hishamhm:
Thank you for the heads-up — the generated headers were out of date. I regenerated all of them.

PR #299, opened 2015-11-01
OpenBSD port updates and error exit improvements

mmcco:
I've occasionally tweaked my OpenBSD port, so this includes a few diffs related to that.

In the process, I noticed that exiting without calling `CRT_done()` corrupts the terminal. So, I also converted all calls to `err.h` functions. We should do a second pass soon looking for all calls to `exit()`.

This changes behavior slightly by making all error exit codes `2`. I doubt that matters, though. Let me know if it might.


hishamhm:
Thank you. For now I think only having the zero/nonzero distinction for error codes is enough. (Being an interactive application, and not one designed for scripting, there's not much point in having detailed error codes.)

PR #303, opened 2015-11-04
match load averages description to what it actually shows

kaefer:
in htop setup of the meters there is a short description to every meter. 
I noticed the load average meter shows the 1 min average, 5 min average and 15 min average. 
In the description it says "15 minutes, 5 minutes, 1 minute". 
Therefore I changed this description to actually match the data which is shown.


hishamhm:
Thank you!

PR #304, opened 2015-11-04
Manpage update

### LINKED ISSUE #297: Description of DATA is wrong
In this ticket ( https://bugzilla.kernel.org/show_bug.cgi?id=101341 ) it got figured out that the description of DATA in the manpage seems to be wrong. The last post from me in this tickets explains the reason:

>Thanks for the information. I think I figured out the problem: The manpage of htop descibes the DATA column as the data segment size + stack size while the manpage of top descibes it as anything except the code size. I'm reporting it then to htop and marking this ticket as invalid.

Sworddragon:
The first commit fixes some typos and the second commit enhances some column descriptions and making them a bit more consistent with other similar sounding ones. Also the second commit fixes #297 by describing DATA as data + stack instead of data segment + stack.

### DIFF
--- a/htop.1.in
+++ b/htop.1.in
@@ -246,21 +246,22 @@ The time the process was started.
 The ID of the CPU the process last executed on.
 .TP
 .B M_SIZE (VIRT)
-Size in memory of the total program size.
+The size of the virtual memory of the process.
 .TP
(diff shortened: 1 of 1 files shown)


hishamhm:
Thank you! :)

PR #308, opened 2015-11-16
Add Darwin swap meter

SaltwaterC:
The 5yo htop-osx fork that I'm still using has a swap meter. Now htop has one too.


hishamhm:
Thank you!

SaltwaterC:
Cheers!

PR #313, opened 2015-11-29
Update calculation of used vs. free memory.

hishamhm:
Account unreclaimable slab and shmem as used memory,
reclaimable slab as cached memory.

Hopefully this presents a more truthful representation of
available vs. used memory on Linux.
See brndnmtthws/conky#130, #242, #67, #263.

Thanks to @OmegaPhil for the detective work!

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -621,6 +621,9 @@ static bool LinuxProcessList_recurseProcTree(LinuxProcessList* this, const char*
 
 static inline void LinuxProcessList_scanMemoryInfo(ProcessList* this) {
    unsigned long long int swapFree = 0;
+   unsigned long long int slab = 0;
+   unsigned long long int shmem = 0;
+   unsigned long long int sreclaimable = 0;
 
    FILE* file = fopen(PROCMEMINFOFILE, "r");
    if (file == NULL) {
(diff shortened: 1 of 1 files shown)


hishamhm:
@OmegaPhil, does this look correct to you?

OmegaPhil:
Minor thing: Since you're not interested in storing slab, might be better to fetch SUnreclaim instead and use it direct in the line 669 calc.

cachedMem still implicitly includes shmem because it comes direct from Cached.

Isn't the conky repo link supposted to be https://github.com/brndnmtthws/conky/issues/130?

Thanks

OmegaPhil:
Hmm, your usedMem calculation (line 669) - usedMem is still 'this->totalMem - this->freeMem', since this is a simple value based off two supposedly golden figures, the total available RAM, and the RAM that the system has literally done nothing with. Both of these should be correct by default.

The problem we are dealing with is more a case of fixing up attributing the cachedMem part of usedMem.

hishamhm:
Come to think of it, you're right. `pl->usedMem` is all used memory (it is split from the buffers and cache values elsewhere). So, I guess the proper calculations are, then:

```c
   this->usedMem = this->totalMem - this->freeMem;
   this->cachedMem = this->cachedMem + sreclaimable - shmem;
   this->usedSwap = this->totalSwap - swapFree;
```

OmegaPhil:
Yep, looks fine.

hishamhm:
Thank you!

OmegaPhil:
Just a heads up, but the shmem issue has come back on the procps mailing list:

http://www.freelists.org/post/procps/OmegaPhilxxxxxxxxxxxxx-Bug799716-free-considers-cached-to-include-SUnreclaim,7

Jaromir is saying that only Available has any use for determining free memory, although the associated calculations look pretty dodgy.

OmegaPhil:
Have had no response to my last mail about this on the procps list in January, so finishing up - Cached as reported by '/proc/meminfo' is wrong in that it includes 'genuinely used memory', i.e. data that can't be flushed and fetched back from somewhere on demand (Shmem this time).

Supposedly all the tools that treat Cached as discardable data are 'wrong', and in 2009 MemAvailable was added to allow people to work out actually-used memory for the first time(??). This doesn't sound right to me, perhaps they don't want to fix Cached since something they're responsible for relies on the broken behaviour?

Anyway its not my position to dictate what htop should be using for available memory here - the current calculation is fine from my perspective since htop removes shmem from cached already (LinuxProcessList.c:LinuxProcessList_scanMemoryInfo at the bottom), but you might want to move to MemAvailable incase other examples of 'genuinely-used' data pop up in Cached and are subsequently fixed in MemAvailable.

hishamhm:
@OmegaPhil Once again, thank you for the analysis and investigation. We'll keep using the current calculation then for the time being (the less I fiddle with these things the better). If something else comes up in the future, I'll revisit these calculations.

hishamhm:
@OmegaPhil Once again, thank you for the analysis and investigation. We'll keep using the current calculation then for the time being (the less I fiddle with these things the better). If something else comes up in the future, I'll revisit these calculations.

hishamhm:
@OmegaPhil Once again, thank you for the analysis and investigation. We'll keep using the current calculation then for the time being (the less I fiddle with these things the better). If something else comes up in the future, I'll revisit these calculations.

PR #315, opened 2015-12-02
add some security checks when running SUID root

mklein-de:
on Darwin, htop needs to run with root privileges to display information
about other users processes. This commit makes running htop SUID root a
bit more safe.

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -513,12 +513,16 @@ void Process_toggleTag(Process* this) {
 }
 
 bool Process_setPriority(Process* this, int priority) {
-   int old_prio = getpriority(PRIO_PROCESS, this->pid);
-   int err = setpriority(PRIO_PROCESS, this->pid, priority);
-   if (err == 0 && old_prio != getpriority(PRIO_PROCESS, this->pid)) {
-      this->nice = priority;
+   if ( Process_getuid == 0 || Process_getuid == (int) this->st_uid ) {
+      int old_prio = getpriority(PRIO_PROCESS, this->pid);
+      int err = setpriority(PRIO_PROCESS, this->pid, priority);
+      if (err == 0 && old_prio != getpriority(PRIO_PROCESS, this->pid)) {
+         this->nice = priority;
+      }
+      return (err == 0);
    }
-   return (err == 0);
+   else
+      return false;
 }
 
 bool Process_changePriorityBy(Process* this, size_t delta) {
@@ -526,7 +530,8 @@ bool Process_changePriorityBy(Process* this, size_t delta) {
 }
 
 void Process_sendSignal(Process* this, size_t sgn) {
-   kill(this->pid, (int) sgn);
+   if ( Process_getuid == 0 || Process_getuid == (int) this->st_uid )
+      kill(this->pid, (int) sgn);
 }
 
 long Process_pidCompare(const void* v1, const void* v2) {
--- a/Process.c
+++ b/Process.c
@@ -513,16 +513,15 @@ void Process_toggleTag(Process* this) {
 }
 
 bool Process_setPriority(Process* this, int priority) {
-   if ( Process_getuid == 0 || Process_getuid == (int) this->st_uid ) {
-      int old_prio = getpriority(PRIO_PROCESS, this->pid);
-      int err = setpriority(PRIO_PROCESS, this->pid, priority);
-      if (err == 0 && old_prio != getpriority(PRIO_PROCESS, this->pid)) {
-         this->nice = priority;
-      }
-      return (err == 0);
+   uid_t euid = geteuid();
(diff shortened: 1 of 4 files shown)


Process.c
```diff
@@ -513,12 +513,16 @@ void Process_toggleTag(Process* this) {
 }
 
 bool Process_setPriority(Process* this, int priority) {
-   int old_prio = getpriority(PRIO_PROCESS, this->pid);
-   int err = setpriority(PRIO_PROCESS, this->pid, priority);
-   if (err == 0 && old_prio != getpriority(PRIO_PROCESS, this->pid)) {
-      this->nice = priority;
+   if ( Process_getuid == 0 || Process_getuid == (int) this->st_uid ) {
```
hishamhm: Couldn't we just let this silently fail when you lack permissions? This is how I've always did on Linux.

hishamhm:
Thank you! If I'm following correctly, you are dropping permissions when you are writing to the filesystem, right?

Process.c
mklein-de: Hi! Could you elaborate? The code does the same as before, it's just wrapped in a check for uid==euid, as a non-root user should not be allowed to change the priority of other users' processes.

mklein-de:
Yes, this patch is about dropping privileges before writing to the filesystem, and checking the user id when sending signals or trying to change the process priority.

Process.c
hishamhm: My question is whether the wrapper is needed. Without the wrapping `if`, htop will try to change permissions, but for non-root users it will simply fail. I'd rather keep it without the wrapping `if`, because the reason why setpriority fails or doesn't fail may have to do with more variables in different OSes (capabilities, etc.) and not just whether the uid is the user or root.

Process.c
```diff
@@ -526,7 +530,8 @@ bool Process_changePriorityBy(Process* this, size_t delta) {
 }
 
 void Process_sendSignal(Process* this, size_t sgn) {
-   kill(this->pid, (int) sgn);
+   if ( Process_getuid == 0 || Process_getuid == (int) this->st_uid )
```
hishamhm: (Same here, please remove the `if`. The changes in `Settings.c` look fine.)

Process.c
mklein-de: December 7 2015 12:03 AM, "Hisham Muhammad"  wrote:
	In Process.c (https://github.com/hishamhm/htop/pull/315#discussion_r46777780):

	> @@ -513,12 +513,16 @@ void Process_toggleTag(Process* this) { > } > > bool Process_setPriority(Process* this, int priority) { > - int old_prio = getpriority(PRIO_PROCESS, this->pid); > - int err = setpriority(PRIO_PROCESS, this->pid, priority); > - if (err == 0 && old_prio != getpriority(PRIO_PROCESS, this->pid)) { > - this->nice = priority; > + if ( Process_getuid == 0 || Process_getuid == (int) this->st_uid ) { 

	My question is whether the wrapper is needed. Without the wrapping if, htop will try to change permissions, but for non-root users it will simply fail. I'd rather keep it without the wrapping if, because the reason why setpriority fails or doesn't fail may have to do with more variables in different OSes (capabilities, etc.) and not just whether the uid is the user or root.When htop runs SUID root, setpriority does *not* fail for non-root users, so a non-root user can change process priority and send signals to arbitrary processes. I guess the if doesn't make that intention clear enough. It's probably clearer if the "dangerous" code is wrapped between seteuid() calls. What do you think?

hishamhm: The seteuid() option seems better, in case the user does have capabilities to change other users' processes. The rationale is that I think it's not up for the htop code to decide what should and shouldn't be allowed.

One question: what happens with euid if I explicitly run the command with `sudo htop` though? I still want htop to be able to change other users' processes that way.

mklein-de: If you run the command with `sudo htop`, euid is 0, so no changes here.

Settings.c
```diff
@@ -154,7 +154,12 @@ static void readFields(ProcessField* fields, int* flags, const char* line) {
 }
 
 static bool Settings_read(Settings* this, const char* fileName) {
-   FILE* fd = fopen(fileName, "r");
+   FILE* fd;
+   uid_t euid = geteuid();
+
+   seteuid(getuid());
+   fd = fopen(fileName, "w");
```
hishamhm: You changed "r" to "w" here. This is overwriting the htop configuration.

mklein-de:
Hi, the build failure looks unrelated:
fatal: unable to access 'https://github.com/hishamhm/htop.git/': Couldn't resolve host 'github.com'

hishamhm:
Merged, thanks!

mklein-de:
Thank *you* for htop ;-) So much better that top!

PR #319, opened 2015-12-10
Redraw the panel when resizing the window during an incremental search

acatton:
`KEY_RESIZE` wasn't handled by the incremental search. Resulting in this
bug:

  * Set your terminal window to a small size.
  * Press `'\'` to filter the processes
  * Maximize your terminal window
  * The list of processes didn't resize.

This change fixes this bug.

Thank you Julian Andrews (@julianandrews) for finding this bug.


hishamhm:
Thank you!! :)

PR #320, opened 2015-12-13
major meters update

etosan:
With these changes all major meters should be covered and htop is finally completely usable on FreeBSD, without Linux procfs emulation. One less module to load into kernel.

Some values like CPU usage per process are lifted directly from kernel and are not that precise, but for an interactive program it's enough, I believe. User can always invoke native top or ps for more details.

Added bonus is jails support, which now makes inspection of jail heavy machines a breeze. 

I hope the code is up to your standard as I am not a C coder. I am interested in extending a memory meter, but I still don't understand many things.

### DIFF
--- a/freebsd/FreeBSDProcessList.c
+++ b/freebsd/FreeBSDProcessList.c
@@ -43,6 +43,15 @@ typedef struct FreeBSDProcessList_ {
    ProcessList super;
    kvm_t* kd;
 
+   int zfsArcEnabled;
+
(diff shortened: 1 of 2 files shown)


hishamhm:
Awesome! I will take a close look at the patch as soon as I can and get back to you :)

hishamhm:
Merged! Thank you!!

etosan:
Thanks for merge, missed it as I am not here often.

PR #325, opened 2015-12-22
Fix text on function bar when moving meters.

Explorer09:
Before:
[Up]Up [Dn]Down [Lt]Left [Rt]Right [Arrow]Confirm [Enter]Delete [Del]Done

After:
[Up]Up [Dn]Down [Lt]Left [Rt]Right [Enter]Confirm [Del]Delete [Esc]Done

This is just a minor fix on the description text. For usability perspective, having both "Confirm" and "Done"  actions can be confusing when a Meter is selected to move. I will suggest removing the [Esc]Done key action or change to something intuitive, such as "Cancel" (and undo the move).


hishamhm:
Thanks!

> I will suggest removing the [Esc]Done key action or change to something intuitive, such as "Cancel" (and undo the move).

Either option would be welcome! (I wonder if the "Cancel" approach would add too much code, though...)

PR #327, opened 2015-12-24
[Experimental] Graph meter coloring by items.

Explorer09:
This patch colors the graph meter based on the items of the bar meter.
For example, if most of the CPU power is running on low-priority
processes, then blue will be shown on the graph columns. If most of the
CPU is running on kernel things, then the columns will be red, and so
on.

Update: My personal to-do list for now:
* [DONE] Figure out if there's a way to make the color determination faster. Currently it's O(items*graphHeight). Is  there a way to make it O(items+graphHeight)?
* Replace prevItemSums and currentItemSums with something more efficient. (Not sure if there is a better structure...)
* Rebase this branch from the current master branch. (This includes adopting the CLAMP macro.)
* [DONE] Fix memory leak on currentItemSums and prevItemSums .

I also need people to review it, ensuring that there are no problems in variables naming, data structures or memory allocation routines. (This is the first time I try such a big change in htop.)

screenshot from 2015-12-24 17 35 01

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -421,7 +421,6 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
       double *prevItemSums = data->prevItemSums;
       double *currentItemSums = data->currentItemSums;
       
-      currentItemSums[0] = this->values[0];
       for (int i = 0; i < items; i++) {
          prevItemSums[i] = currentItemSums[i];
          currentItemSums[i] = ((i > 0) ? currentItemSums[i-1] : 0.0) + this->values[i];
--- a/Meter.c
+++ b/Meter.c
@@ -451,9 +451,9 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
    }
    
    for (int i = nValues - (w*2) + 2, k = 0; i < nValues; i+=2, k++) {
-      const double dot = (1.0 / (GraphMeterMode_pixPerRow * GRAPH_HEIGHT));
-      int v1 = MIN(GraphMeterMode_pixPerRow * GRAPH_HEIGHT, MAX(1, data->values[i] / dot));
-      int v2 = MIN(GraphMeterMode_pixPerRow * GRAPH_HEIGHT, MAX(1, data->values[i+1] / dot));
+      int pix = GraphMeterMode_pixPerRow * GRAPH_HEIGHT;
+      int v1 = MIN(pix, MAX(1, data->values[i] * pix));
+      int v2 = MIN(pix, MAX(1, data->values[i+1] * pix));
 
       for (int line = 0; line < GRAPH_HEIGHT; line++) {
          int line1 = MIN(GraphMeterMode_pixPerRow, MAX(0, v1 - (GraphMeterMode_pixPerRow * (GRAPH_HEIGHT - 1 - line))));
--- a/Meter.c
+++ b/Meter.c
@@ -425,7 +425,7 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
       currentItemSums[0] = this->values[0];
       for (int i = 0; i < items; i++) {
          prevItemSums[i] = currentItemSums[i];
-         currentItemSums[i] = ((i > 0) ? currentItemSums[i-1] : 0) + this->values[i];
+         currentItemSums[i] = ((i > 0) ? currentItemSums[i-1] : 0.0) + this->values[i];
       }
       data->values[nValues - 1] = currentItemSums[items - 1] / this->total;
 
@@ -439,8 +439,8 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
              double lowerBound = this->total * (GRAPH_HEIGHT - 1 - line) / GRAPH_HEIGHT;
              area = MAX(lowerBound, MIN(currentItemSums[i], upperBound)) +
                     MAX(lowerBound, MIN(prevItemSums[i], upperBound));
-             area -= MAX(lowerBound, MIN(((i > 0) ? currentItemSums[i-1] : 0), upperBound)) +
-                     MAX(lowerBound, MIN(((i > 0) ? prevItemSums[i-1] : 0), upperBound));
+             area -= MAX(lowerBound, MIN(((i > 0) ? currentItemSums[i-1] : 0.0), upperBound)) +
+                     MAX(lowerBound, MIN(((i > 0) ? prevItemSums[i-1] : 0.0), upperBound));
              if (area > maxArea) {
                 maxArea = area;
                 dominantColor = CRT_colors[Meter_attributes(this)[i]];
--- a/Meter.c
+++ b/Meter.c
@@ -374,7 +374,7 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
       GraphData* data = (GraphData*) this->drawData;
       for (int i = 0; i < nValues; i++) {
          for (int line = 0; line < GRAPH_HEIGHT; line++) {
-            data->colors[i][line] = CRT_colors[BAR_SHADOW];
+            data->colors[i][line] = BAR_SHADOW;
          }
       }
    }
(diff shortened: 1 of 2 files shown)


Explorer09:
Newer screenshot (with the gray cells problem fixed)
screenshot from 2015-12-24 17 35 01

hishamhm:
This looks _really_ interesting... I do wonder if people would get confused by the fact that the dots and colors work at different resolutions, though.

eworm-de:
I really do like this. Great work!

However depending on the load this looks like a blinking X-mas tree:


No chance to get around this, no?

Explorer09:
@eworm-de 
The blinking is by design. Because one character consists of two columns of data, you know. :)
One way to get around this is to hack the code _not_ to have two columns of data in one character, and instead show one character (two columns of Braille dots) for one data. This would be another patch, and maybe you can conbine this "display mode" with the --no-unicode runtime option you proposed.

eworm-de:
I like the higher resolution with braille characters. ;)
Another way to work around this could be to update graphical meter every second run. This would not change the combination of values.

Would look like this:

Explorer09:
@eworm-de  
Using 2-second update interval is also a good idea.
But the gray dots on your picture...that shouldn't happen. Could you please try the version after commit a94b04665fcf4235dc9ffc96fdc6534c4564e799 ?

eworm-de:
Everything is just fine. ;)
Gray is IO-WAIT, which really happens with heavy IO on LVM snapshot on SAN storage. 
This is an active database server taking a backup in background.

Explorer09:
@eworm-de 
Oh well, that would be another bug... It's a color collision!
In my code, I expect only BAR_SHADOW to be gray, and the gray colors on the cells would indicate they have no data (such as 0% CPU usage or zero use of swap or no laptop battery installed).
My screenshot should have demonstrated that: the left few columns of 'Avg' and 'Mem' are zero data and so are colored gray.

eworm-de:
Hacked some changes for 2-second update interval:
https://github.com/eworm-de/htop/commit/8bfb7b20

Not sure if this is the best solution, but it demonstrates the behaviour pretty well.

hishamhm:
Fixing the blinking updating every 2 readings at once seems the best approach, indeed!

Explorer09:
I need to think of this a little bit...
If you look carefully in my data structures design, you'll see that I don't store the values per item every second, but only the "rasterized" color results. With 2-seconds interval, this means that half of the "rasterized" color in GraphData.colors[x][y] will be wasted. I'm not sure whether you're going to make 2-second interval a default, or just a changeable option.

If it will be an unchangeable default, than you're better not waste cells by only malloc'ing half of the table, and change the use code so that only half of it is used. So, will it be a default or an option?

Explorer09:
And one more thing before you all complain about the blinking in the graph...

@eworm-de 's image has really **unusual** blinking. I've highlighted some rows to explain.
Ideally, if a "cell" has one column full of data, and the other column empty, that column on the cell should be colored the same even when the graph are shifted one column left.
    Correct:
    +---+---+      +---+---+
    |  B|  G|   \  |B  |G  |
    |  B|  G| ---\ |B  |G  |
    |  B|  G| ---/ |B  |G  |
    |  B|  G|   /  |B  |G  |
    +---+---+      +---+---+
    But what I see on @eworm-de 's picture is this:
    +---+---+      +---+---+
    |  B|  G|   \  |G  |B  |
    |  B|  G| ---\ |G  |B  |
    |  B|  G| ---/ |G  |B  |
    |  B|  G|   /  |G  |B  |
    +---+---+      +---+---+
This is nothing to do with IO-wait or the gray cells. It indicated there was something wrong with the coloring. For now, I cannot reproduce such problem. I am not sure whether this problem is caused by the array overflow in `data->prevItemSums` , or by something else. Do you guys tried again and see this weird blinking, especially with commit a94b04665fcf4235dc9ffc96fdc6534c4564e799 ? Please?

eworm-de:
My current branch includes eworm-de@2e71c5c96ba as HEAD:
```
commit 2e71c5c96ba4caea94bef38e73a4f1d8d0097716
Author: Explorer09 <explorer09@gmail.com>
Date:   Wed Jan 6 00:40:17 2016 +0800

    Fix a potential array overflow.
```
The behaviour is still the same...

Explorer09:
@eworm-de 
Weird. It seems like there are still bugs in my algorithm....

eworm-de:
Does it help if I dump some values to a text file for debugging?

Explorer09:
@eworm-de 
Well, is that how you debug in htop? I don't know, but it seems that I found a way to reproduce the problem. Thanks.
With the `stress` program: `nice -n 2 stress -t 1 -c 1`

Explorer09:
@eworm-de Please try commit 95822f3295aecfc8ad4552bcaf4bcd417b87c353 . :)

Explorer09:
My personal to-do list for now:
* Figure out if there's a way to make the color determination faster. Currently it's O(items*graphHeight). Is  there a way to make it O(items+graphHeight)?
* Replace prevItemSums and currentItemSums with something more efficient.

eworm-de:
> Well, is that how you debug in htop?

No, it is not. But I can not give you access to the system with this specific workload. :-p

> Please try commit 95822f3 . :)

Already compiled, will give it a try later.

eworm-de:
Looks better I think.

Explorer09:
Before I continue to write code for this, I ask for opinions about this one more visual change:
Here I wish to distinguish columns with no data / zero values from columns with nothing but a little (<6.2%) IO-wait. And so I add underscores.
(Attach picture:)
screenshot from 2016-01-20 09 41 01
What would you think of the graph on the second window (htop-b)? Will it look better or worse? By the way, if you wish to try the change yourself, I have the [graph-underscore] (https://github.com/Explorer09/htop/tree/graph-underscore) branch for you. Just note that the code is not final.

hishamhm:
@Explorer09 I think the underscore will look misaligned in many fonts (including the one I use in my terminal).

The graph view is already giving a "ballpark" view of numbers and colors; I don't think it needs to (or can) aim for precision like that. The colors will be already less precise than that.

> Figure out if there's a way to make the color determination faster. Currently it's O(items*graphHeight). Is there a way to make it O(items+graphHeight)?

I took the fun challenge and wrote it, it's in branch linear-graph-scan. But according to my rough measurement (comparing how process times of your version and mine progress), it's actually slower than your code. I think the _n_ is too low, so constant factors take over.

Explorer09:
@hishamhm 
a) That's okay. I can get over with no underscores.
b) And you do catch the idea of what I mean by O(items+graphHeight), but I'm not expecting the loop to be so complex. :) Would you allow me to take the challenge then?
By the way, I will keep with my variable naming in my patch, although you are welcome to change it after that.

hishamhm:
> b) And you do catch the idea of what I mean by O(items+graphHeight), but I'm not expecting the loop to be so complex. :) Would you allow me to take the challenge then?

Yeah, I went with the simplest linear flow I could think of. It's not that complex, it just turned out long. It should be possible to make it in a smarter way and improve the performance.

> By the way, I will keep with my variable naming in my patch, although you are welcome to change it after that.

Sure. (Note your original loop (lines 435-445) was over-indented, though.)

Explorer09:
Okay, would you please try mine then? [Explorer09:linear-graph-scan](https://github.com/Explorer09/htop/tree/linear-graph-scan)

Explorer09:
Oops! I gotta add one more thing to todo:
I forgot the deallocation routines when I added `currentItemSums` and `prevItemSums` into the code, and now it leaks memory when GraphData is free'd out.

However, from what I understand about the OOP model in htop, I couldn't figure out how GraphData is destructed, and so where I should put the `free(currentItemSums)` for this. I would need some assistance for this.

eworm-de:
Definitely loving this. :D
Any chance to rebase this on current master?

Explorer09:
@eworm-de
Sure I could. But I wish that the deallocation routines of `currentItemSums` and `prevItemSums` be ready first. Would you help?

PR #331, opened 2016-01-02
OpenBSD fixes and updates

mmcco:
I forgot how awful the process name logic was. It was an initial hack to
get it running, and I forgot to clean it up.

I also had to change a few includes and error function uses.

### DIFF
--- a/openbsd/OpenBSDProcessList.c
+++ b/openbsd/OpenBSDProcessList.c
@@ -60,7 +60,7 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidWhiteList, ui
 
    size = sizeof(fscale);
    if (sysctl(fmib, 2, &fscale, &size, NULL, 0) < 0)
-     errx(1, "fscale sysctl call failed");
+     err(1, "fscale sysctl call failed");
 
    for (i = 0; i < pl->cpuCount; i++) {
       fpl->cpus[i].totalTime = 1;
@@ -90,7 +90,7 @@ static inline void OpenBSDProcessList_scanMemoryInfo(ProcessList* pl) {
    size_t size = sizeof(uvmexp);
 
    if (sysctl(uvmexp_mib, 2, &uvmexp, &size, NULL, 0) < 0) {
-      errx(1, "uvmexp sysctl call failed");
+      err(1, "uvmexp sysctl call failed");
    }
 
    //kb_pagesize = uvmexp.pagesize / 1024;
--- a/openbsd/OpenBSDProcessList.c
+++ b/openbsd/OpenBSDProcessList.c
@@ -130,28 +130,34 @@ char *OpenBSDProcessList_readProcessName(kvm_t* kd, struct kinfo_proc* kproc, in
    size_t cpsz, len = 0, n;
    int i;
 
+   /*
+    * We attempt to fall back to just the command name (argv[0]) if we
+    * fail to construct the full command at any point.
+    */
    arg = kvm_getargv(kd, kproc, 500);
    if (arg == NULL) {
-      // the FreeBSD port uses ki_comm, but we don't have it
-      //return strndup(kproc->ki_comm);
-      if ((s = strdup("[zombie]")) == NULL) {
-	      err(1, NULL);
+      if ((s = strdup(kproc->p_comm)) == NULL) {
+         err(1, NULL);
       }
       return s;
    }
    for (i = 0; arg[i] != NULL; i++) {
       len += strlen(arg[i]) + 1;
    }
-   if ((buf = s = malloc(len)) == NULL)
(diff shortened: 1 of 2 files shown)


openbsd/OpenBSDProcessList.c
```diff
@@ -59,7 +60,7 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidWhiteList, ui
 
    size = sizeof(fscale);
    if (sysctl(fmib, 2, &fscale, &size, NULL, 0) < 0)
-     CRT_fatalError("fscale sysctl call failed");
+     errx(1, "fscale sysctl call failed");
```
Pyrohh: `sysctl(3)` on OpenBSD seems to have a lot of possible failure conditions; why not use `err(3)` instead?

openbsd/OpenBSDProcessList.c
```diff
@@ -89,7 +90,7 @@ static inline void OpenBSDProcessList_scanMemoryInfo(ProcessList* pl) {
    size_t size = sizeof(uvmexp);
 
    if (sysctl(uvmexp_mib, 2, &uvmexp, &size, NULL, 0) < 0) {
-      CRT_fatalError("uvmexp sysctl call failed");
+      errx(1, "uvmexp sysctl call failed");
```
Pyrohh: `sysctl(3)` on OpenBSD seems to have a lot of possible failure conditions; why not use `err(3)` instead?

openbsd/OpenBSDProcessList.c
```diff
@@ -125,34 +126,31 @@ static inline void OpenBSDProcessList_scanMemoryInfo(ProcessList* pl) {
...
+   arg = kvm_getargv(kd, kproc, 500);
+   if (arg == NULL) {
+      return "[zombie]";
+      // the FreeBSD port uses ki_comm, but we don't have it
+      //return strndup(kproc->ki_comm);
+   }
+   for (i = 0; arg[i] != NULL; i++) {
+      len += strlen(arg[i]) + 1;
+   }
+   buf = s = malloc(len);
```
Pyrohh: missing `if (buf == NULL) err(1, NULL` ?

Pyrohh:
Just tested this now and it seems to work fine; I'll report back later any issues I might find.

Pyrohh:
I'm not sure if it's related to this PR, but I got a few SIGABRTS in a row while running `make index` in `/usr/ports` with the error message `bogus pointer (double free?)`. I didn't look into it too deeply, but the backtrace showed it was Happening in `Process_done()` call, so I made the following change and the diff subsided:

```diff
diff --git a/openbsd/OpenBSDProcess.c b/openbsd/OpenBSDProcess.c
index 65caf72..0a630c2 100644
--- a/openbsd/OpenBSDProcess.c
+++ b/openbsd/OpenBSDProcess.c
@@ -196,6 +196,7 @@ void Process_delete(Object* cast) {
    OpenBSDProcess* this = (OpenBSDProcess*) cast;
    Process_done((Process*)cast);
    free(this);
+   this = NULL;
 }
 
 void OpenBSDProcess_writeField(Process* this, RichString* str, ProcessField field) {

```

mmcco:
Hm. There's actually a similar issue that I realized after committing. I had to leave earlier and didn't have a chance to fix it. Basically, these process names get freed eventually, and `"[zombie]"` is static, so we need to use `strdup("[zombie]")` in order to have dynamically allocated memory.

Maybe nulling the pointer like you did fixes an unrelated issue. Or maybe you were running into this and didn't hit it again by chance after adding the null assignment.

hishamhm:
@Pyrohh `this = NULL` in a local variable that is no longer used is not a real fix to the problem. It just happened to change memory positions so that the random bug stopped ocurring. Does Valgrind run on OpenBSD? It could help diagnosing the issue.

Pyrohh:
> this = NULL in a local variable that is no longer used is not a real fix to the problem

Yeah, the patch was wrong; I meant to do `cast = NULL` to prevent a double free, assuming the diagnostic message I got above was correct.

> Does Valgrind run on OpenBSD?

Yes, I'll try it out if I run into the issue again.

mmcco:
I'm running it now as well.

mmcco:
Hmm, I ran it for about 1.5 hours and it only found a single 64-byte leak. Nothing else. IIRC, I still occasionally hit memory-related crashes, though.

However, those could also be platform-independent htop bugs; I'm running with really strong [`malloc.conf`](http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man5/malloc.conf.5) memory sanitization settings, which often exposes bugs that don't surface on other platforms. Even the default settings sometimes do that.

mmcco:
This turned up an unrelated memory leak (see #334).

PR #334, opened 2016-01-03
Plug mem leak, improve CPU enumeration logic

mmcco:
I think this leak may still exist in the FreeBSD port.


hishamhm:
Thanks! I also just noticed the OpenBSD backend still uses the variable name `fpl` (as in FreeBSDProcessList). Perhaps it should be changed to `opl`?

PR #341, opened 2016-01-06
Envscreen

mklein-de:
Pressing the 'e' key displays the (initial) environment of the currently selected process (Linux+Darwin)


hishamhm:
Oh, that's a quite interesting feature! 

Looking at the patch, though, I realize just how much code duplication there is between TraceScreen and OpenFilesScreen. I'm now thinking there should be some sort of InfoScreen "class" that those three extend.

PR #342, opened 2016-01-06
Vikeys

mklein-de:
New setup option: Vi-like cursor movement (off by default)
- j/k moves cursor up/down
- g/G moves cursor to top/bottom

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -47,6 +47,7 @@ typedef enum {
    HTOP_QUIT = 0x10,
    HTOP_REDRAW_BAR = 0x20,
    HTOP_UPDATE_PANELHDR = 0x41, // implies HTOP_REFRESH
+   HTOP_SET_BINDINGS = 0x80,
 } Htop_Reaction;
 
 typedef Htop_Reaction (*Htop_Action)();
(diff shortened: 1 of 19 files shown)


Explorer09:
No, please don't.  Both lowercase 'j' and 'k' have their use in top. It's a bad idea to overload them with another meaning. For example, I have already get used to using 'k' for killing processes in top, then I find out one day 'k' scrolls the list instead because of this damn option...

I have no idea why you definitely need this vi-like movement, don't you have arrow keys on your keyboard? If you don't, then why do you expect there is an F9 key to kill processes?

mklein-de:
It's an option, and it's off by default. So why are you getting so excited about this?
I'm already used to j/k moving the cursor down/up in my preferred editor and, thanks to [pentadactyl](http://5digits.org/pentadactyl/), in my browser.
Why should I move my hand to the arrow keys when I have j/k at a much more convenient place?

Explorer09:
The main problem is that most documentation of 'top' has already established that 'k' is for kill and something like that. And an option in runtime will bring nothing but confusion, especially for new users. I'm really against this _in runtime_. But if you really like such controls, I suggest two approaches:

1. Make it a _compile-time_ option and not a _runtime_ one. That way users who use h/j/k/l as default for cursor movement will know what they are doing. (As help documents and man pages won't reflect to such change.)
2. Implement as Alt+{h,j,k,l} . This is how it's implemented in top for Linux. (http://man7.org/linux/man-pages/man1/top.1.html) And this way users will get least surprise.

mklein-de:
Ok, the second approach makes sense. I'll look into this.

hishamhm:
@Explorer09 Thank you for the input! You verbalized my concerns better than I could have.

@mklein-de I like the second approach (Alt+{h,j,k,l}) too: I think it can be "always on" with no need for a config option, and will result in a much smaller/less intrusive patch.

PR #348, opened 2016-01-13
Check for allocation failure

mmcco:


Explorer09:
Um... You better read these rejected pull requests: #232 #234 
It seems that htop do not check for memory allocation errors for a reason.
@hishamhm I wonder if you should add a comment somewhere in these malloc code stating that it is intentional not to check their return code, so that other developers won't see this as coding mistakes and file such pull requests again and again....
Example comment (I believe this is concise enough):
    /* htop never check for malloc errors. We'll let htop crash upon
   access of the memory rather then adding useless check code. */

mmcco:
Why not just add the `xmalloc` family of functions? They're only a few lines each, and they offer a bunch of benefits over not checking allocations. Dereferencing a null pointer is undefined behavior, and cryptic segfaults leave users unsure about whether they hit a bug or a failed allocation.

hishamhm:
@mmcco switching to something like `xmalloc` is an interesting option, indeed

mmcco:
I'll try to get a draft soon. This is a busy week for me, but I may have a few hours.

Ideally we'd call our curses error function on allocation failure, but I expect that it does allocation of its own. We may have to just call `err()` and accept that the terminal will be corrupted. That isn't a regression from the current state of affairs, anyway.

hishamhm:
Closing this one as we went with another implementation.

Issue #340, opened 2016-01-06
Colors CPU_IOWAIT and BAR_SHADOW collide in default scheme


Explorer09:
I report this issue according to tests in pull request #327.

In the default color scheme, the IO-WAIT color in CPU Meter and the BAR_SHADOW (that is, the text color on the bar where no data is filled on it) are all gray. This will not only confuse my "colored graph" in pull request #327, but also the bar meter mode as well. (Imagine, that when up to 90% of the CPU power is doing IOWAIT, can you see the IO-WAIT color covering up the "90.X%" text on the bar?)

Only in the default color scheme that I discover this problem. Other schemes (except Monochrome, in which color problems may be tolerated) are fine.

May I propose the IO-WAIT color be changed to silver (or non-bold white in some terminals) to avoid this problem?

(It's a rather simple change in CRT.c so I won't file a pull request for this.)

Explorer09:
Attach a screenshot to give you all a better idea of what I mean:
![screenshot from 2016-01-06 13 37 00](https://cloud.githubusercontent.com/assets/2487263/12135778/025edf6c-b47b-11e5-95ab-11ab4d87e699.png)
Note: This I/O load is made up, by a stress test program named `stress` (by Amos Waterland https://packages.debian.org/stable/stress)

hishamhm:
I/O wait is not really CPU "usage", but rather a wait, so that's why it uses the shadow color. I think the gray color is best at depicting that the CPU is not at work.

I understand what you mean when I/O wait is 90+% and you can't tell visually whether it's 95 or 100 because the color matches that of the number. But then, the visual bars are already an approximation, and I think switching I/O wait to white would be confusing.

It's a matter of weighing two conflicting design concerns, and I'll stand by the current design. So, sorry, I think I'll close this as a "wontfix".

PR #355, opened 2016-01-29
Graph meter coloring by items. (v4. Rebased on current master.)

Explorer09:
UPDATE: I'm closing this pull request because of #487. And I'm not willing to maintain this graph-color-new branch for now.
At least one issue is known in this branch:
* The graph coloring (specifically, the area calculation) will go slightly wrong when a meter's "total" value changes. The algorithm does not track the changes of this variable. As a real world example, think of when user changes the swap size (via swapon/swapoff) when a htop process is running. This issue is addressed in #487.

--------

I will close pull request #327 because of this new one. :)

screenshot from 2015-12-24 17 35 01

### DIFF
--- a/DisplayOptionsPanel.c
+++ b/DisplayOptionsPanel.c
@@ -97,5 +97,6 @@ DisplayOptionsPanel* DisplayOptionsPanel_new(Settings* settings, ScreenManager* 
    Panel_add(super, (Object*) CheckItem_newByRef(xStrdup("Count CPUs from 0 instead of 1"), &(settings->countCPUsFromZero)));
    Panel_add(super, (Object*) CheckItem_newByRef(xStrdup("Update process names on every refresh"), &(settings->updateProcessNames)));
    Panel_add(super, (Object*) CheckItem_newByRef(xStrdup("Add guest time in CPU meter percentage"), &(settings->accountGuestInCPUMeter)));
+   Panel_add(super, (Object*) CheckItem_newByRef(xStrdup("Align graph meter bars to reducing color blinking"), &(settings->alignGraphMeter)));
    return this;
 }
--- a/Meter.c
+++ b/Meter.c
@@ -425,14 +425,30 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
       data->values[nValues - 1] = valuesBuf2[items] / this->total;
 
       // Determine the dominant color of the cell in the graph
-      // O(GRAPH_HEIGHT * items)
+      // O(GRAPH_HEIGHT + items) (linear time)
+      int buf1Start = 0, buf2Start = 0;
       double low, high = 0.0;
       for (int h = 0; h < GRAPH_HEIGHT; h++) {
          low = high; // low == (this->total * (h) / GRAPH_HEIGHT);
          high = this->total * (h + 1) / GRAPH_HEIGHT;
          double maxArea = 0.0;
          int color = BAR_SHADOW;
-         for (int i = 0; i < items; i++) {
+         for (int i = MIN(buf1Start, buf2Start); ; i++) {
+            if (valuesBuf1[i] < high)
+               buf1Start = i;
+            if (valuesBuf2[i] < high)
+               buf2Start = i;
+            if (i >= items)
+               break; // No more items
+            if (valuesBuf1[i] >= high && valuesBuf2[i] >= high) {
+               // This cell is finished. Rest of values are out of this.
+               break;
+            }
+            // Skip items that have no area in this cell.
+            if (valuesBuf1[i] >= high)
+               i = MAX(i, buf2Start);
+            if (valuesBuf2[i] >= high)
+               i = MAX(i, buf1Start);
             double area;
             area = CLAMP(valuesBuf1[i + 1], low, high) +
                    CLAMP(valuesBuf2[i + 1], low, high);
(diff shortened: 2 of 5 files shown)


hishamhm:
Awesome! It does beat my version in CPU usage. I do think we have to switch this to move only on every two readings, so avoid the blinking effect.

eworm-de:
I think I would be happy with it as is. It still blinks a bit, but eye cancer was caused by a bug that @Explorer09 fixed.

Explorer09:
@hishamhm You said that. Here you go:
5a767abb7208addfa7a9c97df7703aff75544434

hishamhm:
Thank you! There's so many people who email asking "why are the numbers inside the meters sometimes gray and sometimes red?" that I just _know_ that any kind of blinking would get a lot of people confused.

Explorer09:
@hishamhm  Email? You mean there are more people than us three testing this graph meter?
By the way, as I said in the commit description, I would like the behavior be tweakable in the Setup screen.

hishamhm:
@Explorer09 oh, I meant the feedback I got about the regular bar meters over the years. In any case, regardless of whether it's tweakable or not, tuning proper defaults is of prime importance — in my experience, the vast majority of users never goes past the defaults.

eworm-de:
But the default configuration does not include graphical meters at all, no?

hishamhm:
@eworm-de [Graph] mode no, but [Bar] yes.

Explorer09:
@hishamhm 
> I meant the feedback I got about the regular bar meters over the years. In any case, regardless of whether it's tweakable or not, tuning proper defaults is of prime importance — in my experience, the vast majority of users never goes past the defaults.

I'm sorry. If I were the first time htop user, I would complain that bar meter behavior too. It's not so intuitive in text mode, and more confusing in monochrome. (Remember #340 that I have reported to you? It's related.)
Most people expects this intuitive behavior:
`Avg[||||||||||||||||||||||||]100.0%`
`Mem[|||||||||||||||||||]4.24G/7.71G`

`Avg[|||########****.........]100.0%`
`Mem[|||||||||##********]4.24G/7.71G`

But this is htop's behavior:
`Avg[||||||||||||||||||||||||100.0%]`
`Mem[|||||||||||||||||||4.24G/7.71G]`

`Avg[|||###########*****.....100.0%]`
`Mem[||||||||||||||##***4.24G/7.71G]`

I hope you can sense the difference easily.

hishamhm:
@Explorer09 yes, it's bad in monochrome, I know. But I didn't think that scenario is common enaugh to warrant that much special casing.

A heads-up: I'm making the final touches to release htop 2.0. I don't think this PR will make it (the current PR did crash on me once and there are still visual things I want to tweak about this), but it will definitely be in the release after this (which definitely won't take as long).

Explorer09:
@hishamhm 
Just a reminder that you may cherry-pick b515824fca036708d72ca5bcf19fad7e51953497 out to your master branch. It fixes a crash that should also happen on your current master branch, without my graph-color feature.

hishamhm:
@Explorer09 thanks for the heads up, much appreciated. I haven't been able to finish the package yet (still travelling: left Belgium, now in Portugal; will be back home in Brazil tomorrow night only) but I'm set on making this release as fast as I can.

Explorer09:
V3! Rebased again on current master (post-2.0.0)

eworm-de:
The most recent value is still blinking - not with colour, it does not show every second turn.

eworm-de:
Hmm, looking at this for some time... Perhaps we could define this to be the correct behaviour. :-p

Explorer09:
@eworm-de 
Yes, my patch does make the align_graph_meter the default behavior! In fact, I'm using this version right now. (But have manually disabled align_graph_meter as I prefer the non-aligned display.) :)
And for your "most recent value blinking" complaint -- just watch carefully please. It's the result of aligning. I'm avoiding the 2-interval updating approach -- that just doesn't look good here.

eworm-de:
You do the following:

* collect even data
* update the graph
* collect odd data
* update the last column of graph

Instead you could:

* collect even data
* do nothing...
* collect odd data
* update the graph

This way you are behind for even data, but completely avoid blinking.

Explorer09:
@eworm-de I don't want the graph to appear less responsive when aligning is enabled. And I said _reducing_ color blinking, and not avoiding it completely.

And I believe this discussion is going political now. :(

hishamhm:
Thank you for the input, guys! I can't test this today but I'll give a good close look at this later and then make the call on the final behavior.

hishamhm:
Marking this for htop 2.1; I still may tweak the final behavior but some form of colored graph is definitely in. :) Thank you @Explorer09 for the hard work on this!

Explorer09:
Discovered a bug with this! Don't merge to master branch yet until I fix this.
For some meter classes (e.g. `SwapMeter`, `LoadMeter`), their `maxItems` properties are undefined.
However, my `Meter_getMaxItems()` macro does not assume the default value 1 and instead will get the value 0 instead, which in turn will corrupt the allocated memory of `drawData` in the graph drawing routine (and crash).

Although I can add a conditional in `Meter_getMaxItems` to check for value 0, will it be better if we enforce the `maxItems` property of ALL meter classes so that we can save the conditional code in runtime? (The conditionals can become assertions for debug build)

Explorer09:
Closing this pull request because of #487. And I'm not willing to maintain this graph-color-new branch for now.
Note a known issue here:
* The graph coloring (specifically, the area calculation) will go slightly wrong when a meter's "total" value changes. The algorithm does not track the changes of this variable. As a real world example, think of when user changes the swap size (via swapon/swapoff) when a htop process is running.

PR #356, opened 2016-01-31
Add xmalloc support

mmcco:
I took xmalloc.c and xmalloc.h from OpenSSH Portable and modified them
as needed. The coding style is different from htop's, but I'd suggest
keeping local modification to minimum to make merging easier.

We still need to decide whether we're going to add second variants to
call CRT_fatalError() instead of errx(). This would allow us to exit on
error without corrupting the terminal.

However, most allocations in htop assume success and crash on failure.
Using xmalloc et al. in these cases is an unambiguous improvement.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -24,7 +24,7 @@ BatteryMeter.c Process.c ProcessList.c RichString.c ScreenManager.c Settings.c \
 SignalsPanel.c StringUtils.c SwapMeter.c TasksMeter.c UptimeMeter.c \
 TraceScreen.c UsersTable.c Vector.c AvailableColumnsPanel.c AffinityPanel.c \
 HostnameMeter.c OpenFilesScreen.c Affinity.c IncSet.c Action.c EnvScreen.c \
-InfoScreen.c
+InfoScreen.c XMalloc.c
 
 myhtopheaders = AvailableColumnsPanel.h AvailableMetersPanel.h \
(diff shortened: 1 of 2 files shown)


mmcco:
I've tested this and it worked for me. I thought it'd be cleanest to commit the addition and the uses separately, though.

mmcco:
I just realized that`reallocarray`, `strlcpy`, and `vasprintf` need to be either replaced or included from other OpenSSH Portable files. I'll look for the most elegant solution.

mmcco:
It should be fixed now.

I just removed `xreallocarray` for the time being. It's a really nice function though, so I may re-add it later if there's interest. It's a thin wrapper around `malloc`, so it'd be an easy addition.

hishamhm:
Cool, thanks!

I avoided checking mallocs for a long time because it is basically useless on Linux userspace (due to memory overcommit) but now that htop is portable and used in more security-conscious OSes, I won't assume Linux-like behavior like this anymore.

I just wrote a simliar xmalloc implementation yesterday following your suggestion. It's the last thing I want to get in the codebase before htop 2.0. I'll give your PR a run later today and check against my code. I "pre-announced" htop 2.0 in FOSDEM yesterday saying the tarball would be out "Real Soon Now, probably later tonight" so I really want to get this done today.

hishamhm:
I ended up committing the implementation I had. Let me know if there you see any issues with it.

PR #359, opened 2016-02-11
Add FreeBSD battery support

myfreeweb:
Hi, thanks for htop 2.0, this is great :-)

This pull request adds support for the battery meter on FreeBSD.
Tested on both a laptop and a desktop.


hishamhm:
Awesome! Thank you!

PR #363, opened 2016-02-11
Fix process memory on FreeBSD

myfreeweb:
The formulas for memory weren't correct, everything was reported 4x as large :D

See screenshot:
htop

### DIFF
--- a/freebsd/FreeBSDProcessList.c
+++ b/freebsd/FreeBSDProcessList.c
@@ -477,8 +477,8 @@ void ProcessList_goThroughEntries(ProcessList* this) {
       }
 
       // from FreeBSD source /src/usr.bin/top/machine.c
-      proc->m_size = kproc->ki_size / 1024;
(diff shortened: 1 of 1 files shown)


hishamhm:
Thank you!

hishamhm:
Looks like I need to accumulate those platform-specific fixes over the next few days and release htop 2.0.1 soonish :)

myfreeweb:
Yeah...

By the way:

1. `htop` doesn't show threads on FreeBSD, I might look into that soon.
2. The `Tasks` counter always shows the number of, like, all the tasks it could find. (On all platforms.) In FreeBSD's `top`, you can toggle the displaying of threads with `H` and of kernel processes with `S`. The number of tasks changes depending on what toggles are active, e.g. currently on my laptop: nothing – 84, `S` – 107, `H` – 175, `H+S` – 497. And `htop` shows 437 tasks. What do you think about making that number depend on the visibility settings?

PR #365, opened 2016-02-11
Add travis badge, update README

deric:
I've updated compiling instruction for fast skim readers :) Geeks cloning `htop` directly from github, can't use the `./configure` script. Is it ok this way?

P.S. would you mind moving the source code into subfolders, so that the repo would be in more github-like format? Anyway, it's a non issue, and thanks a lot for the awesome htop!

### DIFF
--- a/README
+++ b/README
@@ -1,5 +1,7 @@
 [![Build Status](https://travis-ci.org/hishamhm/htop.svg?branch=master)](https://travis-ci.org/hishamhm/htop)
-htop
+[![PayPal donate](https://img.shields.io/badge/paypal-donate-green.svg)](http://hisham.hm/htop/index.php?page=author#paypal)
+
+[htop](http://hisham.hm/htop/)
 ====
 
 by Hisham Muhammad <hisham@gobolinux.org> (2004 - 2016)
--- a/README
+++ b/README
@@ -41,11 +41,8 @@ probably used to the common routine:
 
     ./autogen.sh && ./configure && make
 
-If you want to replace currently installed version of `htop` use `--prefix` flag. `--prefix=/usr`
-will install `htop` into `/usr/bin/htop`.
-
-    ./autogen.sh && ./configure --prefix=/usr && make
-    make install
+By default `make install` will install into `/usr/local`, for changing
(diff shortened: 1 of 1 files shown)


Explorer09:
I don't believe configuring htop with `--prefix=/usr` on a production system is a good idea. `/usr` is supposed to contain distro packages and be maintained by distro's package managers. You will overwrite something if you are not careful. A better way is to tell the user to remove the distro package first, and then build and install from source.

deric:
Right, I agree. It might depend on Linux distribution but, `/usr/local/bin` precedes `/usr/bin` on `$PATH`. So removing the distribution package might not be necessary.

hishamhm:
Nice, thanks! 

But I wouldn't put instructions for a single distro here in the README (I don't use Debian/Ubuntu and don't wish to maintain those instructions myself, and there are many other downstream distros out there, I don't feel like promoting Debian/Ubuntu in particular, even though they do great work).

Also, ./autogen.sh is necessary when building from Git, but not when building from the tarball, and this README is included in both.

Apart from that, the other changes look good, thank you!

deric:
@hishamhm github should show the `README.md` file by default. If we remove symlink to `README`, you can copy to tar file just the README with slightly different instructions. Is that ok?

deric:
I can write build requirements for a few distributions, just give me a list of those you'd like to be mentioned. It might support the impression, that htop is a actually a multi-platform program :smiley:.

README
```diff
@@ -1,50 +1,56 @@
...
+
+By default `make install` will install into `/usr/local`, for changing
+the path use `./configure --prefix=/some/path`.
+
+See the manual page (`man htop`) or the on-line help ('F1' or 'h'
+inside `htop`) for a list of supported key commands.
 
-When fetching the code from the development repository, you need
-to run the `./autogen.sh` script, which in turn requires autotools
-to be installed.
```
hishamhm: I appreciate having an easily copy-pastable command there, but this sentence is a bit more detailed then than the new version.

hishamhm:
Merging! I'll tweak it further later if necessary :)

PR #366, opened 2016-02-11
ADDED: Toggle Expand and Collapse all functionality with F12 key

sharma-abhishek:
Kindly review expand and collapse all processes (toggle) functionality with F12 key.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -148,6 +148,15 @@ static void tagAllChildren(Panel* panel, Process* parent) {
    }
 }
 
+static bool expandCollapseAll(Panel* panel) {
+   for (int i = 0; i < Panel_size(panel); i++) {
+      Process* p = (Process*) Panel_get(panel, i);
+      if(!p) return false;
+      p->showChildren = !p->showChildren;
+   }
+   return true;
+}
+
 static bool expandCollapse(Panel* panel) {
    Process* p = (Process*) Panel_getSelected(panel);
    if (!p) return false;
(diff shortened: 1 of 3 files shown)


hishamhm:
Hi! Thanks for taking the time to write a PR! But what's the expected behavior? On Linux systems, this will collapse init and basically hide all processes. I'm afraid it doesn't sound too useful...

sharma-abhishek:
Hi hishamnhm,

It is actually an add-on to expand/collapse functionality which basically group all the child/worker processes into single entity (parent process). It will be of great use when the list of child processes are too long to fit in a console and user wants to scroll down to find the process quickly.

gsaga:
Hello @hishamhm , I can see what you're getting at, but I've verified by compiling that this fix **doesn't cause the issue you think it does**. It collapses all processe trees except the init, so I thinks it can be merged into the next release.

https://www.pinterest.com/pin/149955862565400621/

5shekel:
hey @sharma-abhishek , im running your patch and its great. long time request! thanks so much

sharma-abhishek:
Thanks @gsaga and @5shekel. Good to know that it’s working for you guys. Hoping to see this request merged soon.

5shekel:
further usage revealed a rough edge. 
start htop, click F12, all fold (aside from init), all good 

but... in situation when 
1. htop is already opened and folded
2. starting new processes - the new processes will not "auto-fold"
3. and hitting F12 again will not collapse the new guy, but only prior processes (not sure about which ones actually)
tested on arch linux

SriramKid:
> Just tried on Arch Linux and initially nothing change, after few seconds it closes with
> `PID abort (core dumped)`

Same problem here, just compiled and ran htop on a Manjaro install, got "Aborted (core dumped)". Please help D=

PR #376, opened 2016-02-12
Add support for cachedMem and fix usedMem on OpenBSD.

juanfra684:
Tested against htop 2.0 on OpenBSD 5.9 -current. In my tests, the numbers in the meters are quite similar to the output of top (unlike htop, top shows the memory occupied by the cache as used).

The one-line change in OpenBSDProcessList.h is for to avoid to run MakeHeader.py during the build phase.

/cc @mmcco

### DIFF
--- a/openbsd/OpenBSDProcessList.c
+++ b/openbsd/OpenBSDProcessList.c
@@ -10,18 +10,19 @@ in the source distribution for its full text.
 #include "OpenBSDProcessList.h"
 #include "OpenBSDProcess.h"
 
-#include <unistd.h>
-#include <stdlib.h>
 #include <err.h>
 #include <errno.h>
-#include <sys/types.h>
+#include <fcntl.h>
+#include <sys/mount.h>
 #include <sys/param.h>
 #include <sys/proc.h>
-#include <sys/sysctl.h>
-#include <sys/user.h>
-#include <fcntl.h>
-#include <string.h>
 #include <sys/resource.h>
+#include <sys/sysctl.h>
+#include <sys/types.h>
+#include <sys/user.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
 
 /*{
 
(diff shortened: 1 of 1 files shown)


mmcco:
Thanks for this - it looks good. I'll probably have time to test tomorrow. In the meantime, a few quick comments:

* Be sure to adhere to the existing style. It isn't KNF, but... it's something. That includes three-space indents (see line 116).
* It probably isn't necessary to add yourself to the header copyright.   :-)
* I think we should keep the includes alphabetized. That said, they probably need to be rearranged in all files in a separate PR.

juanfra684:
I updated the commit with your feedback. Thanks! :)

It took me hours to get the correct value for `usedMem`. I'm still not 100% convinced.

mmcco:
> It took me hours to get the correct value for usedMem. I'm still not 100% convinced.

Welcome to system monitoring code.  :-)  I try to pull functions from top(1) and ps(1) whenever possible.

openbsd/OpenBSDProcessList.c
```diff
@@ -95,16 +96,29 @@ void ProcessList_delete(ProcessList* this) {
...
+   size_t size_uvmexp = sizeof(uvmexp);
 
-   if (sysctl(uvmexp_mib, 2, &uvmexp, &size, NULL, 0) < 0) {
+   if (sysctl(uvmexp_mib, 2, &uvmexp, &size_uvmexp, NULL, 0) < 0) {
       err(1, "uvmexp sysctl call failed");
    }
 
    //kb_pagesize = uvmexp.pagesize / 1024;
-   pl->usedMem = uvmexp.active * pageSizeKb;
+   //pl->usedMem = uvmexp.active * pageSizeKb;
```
hishamhm: No need to keep old code around as comments; we can always refer to version control to look at the old version. :)

hishamhm:
Thanks for the review @mmcco! Once you give it a spin, let me know and I'll merge.

falsovsky:
I am currently using this, and it seems good.

hishamhm:
Merged, thank you!

mmcco:
Thanks for merging - I forgot to come back to this. I'll test it out now.

PR #380, opened 2016-02-12
Simplify autogen.sh

fingolfin:
Make use of the autoreconf tool

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,15 +1,3 @@
 #!/bin/sh
-
-glibtoolize=$(which glibtoolize 2> /dev/null)
-if [ ${#glibtoolize} -gt 0 ]
-then libtoolize=glibtoolize
-else libtoolize=libtoolize
-fi
-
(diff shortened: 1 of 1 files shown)


hishamhm:
Thanks! But note that the CI build failed. Can you get it to build cleanly? Also, will this work on systems where libtoolize is called glibtoolize?

fingolfin:
That should fix it (I didn't realize that before autoconf 2.69, one had to manually ensure the "m4" subdirectory exists).

Yes, it does work with glibtoolize, which is why it works on that Mac OS X box :-)

Actually, it is a bit more complicated like that: It works correctly provided it has been setup correctly, which should be the case if one uses the autoconf/automake/libtool packages provided with most (hopefull: all) package managers. But I am pretty sure that if you try hard enough, you can arrange things to fail... but then that would be a case of a broken package resp. broken system.

hishamhm:
Cool! Let's give it a try!

PR #391, opened 2016-02-13
Add MEM% for processes in process list on FreeBSD.

### LINKED ISSUE #373: FreeBSD MEM% 0
All percentages in PERCENT_MEM column are always 0%.

FreeBSD 10.2-RELEASE-p9, htop 2.0.0

gaod:
Add MEM% for processes in process list on FreeBSD.

I think this patch should fix #373.


hishamhm:
Merged your fix, thank you!

Issue #379, opened 2016-02-12
Command line options aren't displayed on Mac OS X


klas-genestack:
When I start  `./htop --user $(whoami)`, I can only see `./htop` in *htop*.

Mac OS X 10.11.3, *htop* from current ''master'' (`2.0.0-11-g1bedac1`)

mhinz:
FWIW, it works for me on the same OS version, same commit.

melomac:
Can reproduce, program arguments are missing. Please compare with `ps axu` output.

Tested on OS X 10.11.3 and commit 0b70439316b4e4608c0916317ded7e6e56982de6.

mhinz:
Oops, I misread the issue. I can reproduce this as well. Sorry for the confusion.

hunsakerbn:
The following patch fixes both the basename highlight problem and the process arguments not showing problem.

--- htop-2.0.0/darwin/DarwinProcess.c   2016-02-02 07:58:45.000000000 -0700
+++ htop-2.0.x/darwin/DarwinProcess.c   2016-02-18 14:09:59.000000000 -0700
@@ -286,9 +286,9 @@
       /* The command is from the old Mac htop */
       char *slash;

-      proc->comm = DarwinProcess_getCmdLine(ps, false);
-      slash = strrchr(proc->comm, '/');
-      proc->basenameOffset = (NULL != slash) ? (slash - proc->comm) : 0;
+      proc->comm = DarwinProcess_getCmdLine(ps, true);
+      proc->basenameOffset = strlen(DarwinProcess_getCmdLine(ps, false));
+
    }

    /* Mutable information */

hishamhm:
Thanks @hunsakerbn for confirming this logic fixes the issue! I have a Mac shell currently available and I got a patch together that obtains the same results while calling DarwinProcess_getCmdLine only once.

PR #394, opened 2016-02-13
Show process arguments on OSX

mhinz:
Follow up to: https://github.com/hishamhm/htop/issues/379

Currently calling `DarwinProcess_getCmdLine(ps, false)` disables showing the process arguments. Is there a certain reason for this (unfinished feature, forgot to re-enable after testing)?

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -286,7 +286,7 @@ void DarwinProcess_setFromKInfoProc(Process *proc, struct kinfo_proc *ps, time_t
       /* The command is from the old Mac htop */
       char *slash;
 
-      proc->comm = DarwinProcess_getCmdLine(ps, false);
+      proc->comm = DarwinProcess_getCmdLine(ps, true);
       slash = strrchr(proc->comm, '/');
       proc->basenameOffset = (NULL != slash) ? (slash - proc->comm) : 0;
    }


hishamhm:
I have no idea why is this set to false, and the code in `DarwinProcess_getCmdLine` is quite complicated (and seems to been carried over from ps, into htop-osx, into htop). There's a big commented-out block with `#if 0` ( https://github.com/hishamhm/htop/blob/e9b32eb62f6505fd065022cdd085ae6d8e5e0e46/darwin/DarwinProcess.c#L198-L228 ) and the code doesn't seem to match the preceding comment in https://github.com/hishamhm/htop/blob/e9b32eb62f6505fd065022cdd085ae6d8e5e0e46/darwin/DarwinProcess.c#L170-L197

Looking closely at the code, it seems that a side-effect of arguments being dropped is that the basenameOffset calculation for the `Highlight program "basename"` feature in the Display Options section of Setup. https://github.com/hishamhm/htop/blob/e9b32eb62f6505fd065022cdd085ae6d8e5e0e46/darwin/DarwinProcess.c#L289-L291 

I don't have a Mac to test this myself, so could you help me test this? Please take a look of how the highlighted basename works when your PR is applied. I suspect it will be incorrect. The correct calculation, I believe, would be to record the length of the first token detected by this loop before the '\0' is converted into a space https://github.com/hishamhm/htop/blob/e9b32eb62f6505fd065022cdd085ae6d8e5e0e46/darwin/DarwinProcess.c#L170-L197
This value would be then the proper value for `basenameOffset` (The `show_args` arguments could be dropped altogether and replaced by an `int*` variable so that we could return the basename offset by reference).

Hope this all makes sense! Feel free to ask questions if I sound confusing!

Thank you!!!

melomac:
It indeed partially breaks the _Highlight program "basename"_ option with some processes, exemple `Google Chrome Helper`.

mhinz:
Closed by https://github.com/hishamhm/htop/commit/42c4459375f81d690170c58cbb1078db8321fa30.

PR #397, opened 2016-02-14
linux: use get/setpriority rather than syscall

coypoop:
This shouldn't create a functional change.
I am using the Linux code as-is on NetBSD as it has /proc for compatibility. I only needed this one minor change. It has also worked in previous versions. 
It may be working for SunOS operating systems similarly (as someone has made SunOS specific changes on pkgsrc), but I cannot vouch for it.
Other than values in /proc/pid/statm being set to zero rather than the real value, it works fine.

BTW, the unsupported platform code seems broken.

### DIFF
--- a/linux/LinuxProcess.c
+++ b/linux/LinuxProcess.c
@@ -14,6 +14,7 @@ in the source distribution for its full text.
 #include <stdlib.h>
 #include <unistd.h>
 #include <string.h>
+#include <sys/resource.h>
 #include <sys/syscall.h>
 
 /*{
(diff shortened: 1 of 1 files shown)


hishamhm:
Hi!

Thanks for reporting on NetBSD compatibility!

This patch, however, is incorrect, because the syscall is for setting _IO_ priority (a Linux-specific, non-POSIX feature), not scheduling priority (for which I already use the portable setpriority in Process.c).

I made a different commit to disable this syscall on platforms that don't have it, please let me know if it fixes the build for you.

I also fixed the `unsupported` platform, thanks for the heads up!

PR #414, opened 2016-02-17
Mouse-friendly functions bar for meters panel

Explorer09:
selection_215

Before:

```
SpaceType  EnterMove  DelDeleteEscDone  |
~~~~~      ~~~~~      ~~~      ~~~      |
UpUp     DnDown  LtLeft  RtRight EnterConfirmDelDeleteEscDone  |
~~       ~~      ~~      ~~      ~~~~~       ~~~      ~~~      |
```

After:

```
SpaceType  EnterMove                                           DelDeleteF10Done  |
~~~~~      ~~~~~      ~~                                       ~~~      ~~~      |
SpaceType  EnterLock  UpUp    DnDown  <-Left  ->Right          DelDeleteF10Done  |
~~~~~      ~~~~~      ~~      ~~      ~~      ~~      ~~       ~~~      ~~~      |
```
- Align 'Delete' and 'Done' to the right to match functions on other screens.
  (Accidental clicking is avoided as a side benefit.)
- You could change meter type while in moving mode. New bar now hints this.
- Two Enter key functions are put in the same place and so mouse clicks there
  act like functions toggle. (The wording change to 'Lock' is also to reflect
  this.)
- '<-' and '->' instead of 'Lt' and 'Rt' abbreviation as the latter is not
  widely seen and arrows shapes are obvious. :)
- I'm advocating to avoid mentioning 'Esc' key anywhere in functions bar now,
  because for ncurses apps, Esc key does not take effect immediately, but
  rather in a couple seconds. Double 'Esc' presses does take it immediately,
  but new users unfamiliar with ncurses will not notice it. So here, 'F10',
  which takes effect immediately, is better.

### DIFF
--- a/MetersPanel.c
+++ b/MetersPanel.c
@@ -31,7 +31,9 @@ struct MetersPanel_ {
 
 }*/
 
-static const char* MetersFunctions[] = {"Type  ", "Move  ", "Delete", "Done  ", NULL};
+// Note: In code the meters are known to have bar/text/graph "Modes", but in UI
+// we call them "Styles".
+static const char* MetersFunctions[] = {"Style ", "Move  ", "Delete", "Done  ", NULL};
 static const char* MetersKeys[] = {"Space", "Enter", "Del", "Esc"};
 static int MetersEvents[] = {' ', 13, KEY_DC, 27};
 
--- a/MetersPanel.c
+++ b/MetersPanel.c
@@ -33,13 +33,17 @@ struct MetersPanel_ {
 
 // Note: In code the meters are known to have bar/text/graph "Modes", but in UI
 // we call them "Styles".
-static const char* MetersFunctions[] = {"Style ", "Move  ", "Delete", "Done  ", NULL};
-static const char* MetersKeys[] = {"Space", "Enter", "Del", "Esc"};
-static int MetersEvents[] = {' ', 13, KEY_DC, 27};
+static const char* MetersFunctions[] = {"Style ", "Move  ", "                                       ", "Delete", "Done  ", NULL};
+static const char* MetersKeys[] = {"Space", "Enter", "  ", "Del", "F10"};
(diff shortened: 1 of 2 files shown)


hishamhm:
This is pretty interesting, but I'll hold this off until after htop 2.0.1, which I want to be a strictly bug-fix release (due until the end of the week).

* Since we're rethinking terminology, I think 'Style' is better than 'Type' (I went through this confusion explaining meters to people — "style" is more often understood as meaning presentation (e.g. CSS), while CPU and Memory meters are understood as being "two types of meters")
* Well, if you're using "<-", then why not go all the way and use "←" and friends when UTF8 is available? :)
* I'm not sure about the 'Esc' issue (I think it is more intuitive sitting alongside Enter, Delete and the arrows).

Explorer09:
@hishamhm 
I'm fine with delaying this to next major release. (Along with the color graph meters, of course.)

* The "style" wording is good, but I just have never thought of that. I was addressing different usability problem when I wrote this change.
* The reason of me avoiding UTF-8 left & right arrows (← →) is that I cannot be sure whether they will display full-width or half-width. No terminals or fonts are consistent with this. (Really. I have personal experience: Chinese terminals and BBSes assume arrows to be full-width, and our BBS arts have to rely with that assumption.) If character widths were not important I would use that, but that's not the case.
* For the 'Esc' issue: Yes I believe it's more intuitive, but if I want to make the wording, I would say `EscEsc` on the functions bar to let users aware of such issue.

And just for reference, this is what the menuconfig (the Linux kernel config interface) help text says:

    To exit a menu use the cursor keys to highlight the <Exit> button
    and press <ENTER>.
    
    Shortcut: Press <ESC><ESC> or <E> or <X> if there is no hotkey
              using those letters.  You may press a single <ESC>, but 
              there is a delayed response which you may find annoying.

PR #418, opened 2016-02-19
htop: improve -p error handling

GiedriusS:
A bunch of small changes to improve the situation with -p input handling. First attempt to contribute to htop. Really for maximum correctness we should switch to using `long` for storing PIDs because that is guaranteed to be big enough to store any kind of PID.

### DIFF
--- a/htop.c
+++ b/htop.c
@@ -25,6 +25,8 @@ in the source distribution for its full text.
 #include <string.h>
 #include <time.h>
 #include <unistd.h>
+#include <errno.h>
+#include <limits.h>
 
 //#link m
 
@@ -51,6 +53,25 @@ static void printHelpFlag() {
          "See 'man htop' for more information.\n",
          stdout);
    exit(0);
+}
+
+static bool parsePID(char *pid_str, unsigned int *result) {
+   char *endptr;
+   unsigned long tmp_result;
+
+   if (pid_str[0] == '-')
+      return false;
+
+   errno = 0;
+   tmp_result = strtoul(pid_str, &endptr, 10);
+   if (*endptr != '\0' || errno != 0)
+      return false;
+
+   if (tmp_result > UINT_MAX)
+      return false;
+
+   *result = (unsigned int) tmp_result;
+   return true;
 }
 
 // ----------------------------------------
@@ -137,9 +158,15 @@ static CommandLineSettings parseArguments(int argc, char** argv) {
             }
 
             while(pid) {
-                unsigned int num_pid = atoi(pid);
-                Hashtable_put(flags.pidWhiteList, num_pid, (void *) 1);
-                pid = strtok_r(NULL, ",", &saveptr);
+               unsigned int num_pid;
+
+               if (!parsePID(pid, &num_pid)) {
+                  fprintf(stderr, "Error: invalid PID \"%s\".\n", pid);
+                  exit(1);
+               }
+
+               Hashtable_put(flags.pidWhiteList, num_pid, (void *) 1);
+               pid = strtok_r(NULL, ",", &saveptr);
             }
             free(argCopy);
 


htop.c
```diff
@@ -137,9 +139,33 @@ static CommandLineSettings parseArguments(int argc, char** argv) {
...
+               if (*endptr != '\0' || errno != 0) {
+                  fprintf(stderr, "Error: failed to parse PID \"%s\"", pid);
+                  if (errno == 0)
+                     printf(".\n");
+                  else
+                     printf(": %s.\n", strerror(errno));
+                  goto err_next;
+               }
+
+               if (num_pid > UINT_MAX) {
```
hishamhm: how can strtoul produce a number greater than UINT_MAX?

htop.c
```diff
@@ -137,9 +139,33 @@ static CommandLineSettings parseArguments(int argc, char** argv) {
...
 
             while(pid) {
-                unsigned int num_pid = atoi(pid);
-                Hashtable_put(flags.pidWhiteList, num_pid, (void *) 1);
-                pid = strtok_r(NULL, ",", &saveptr);
+               char *endptr;
+               unsigned long num_pid;
+
+               if (pid[0] == '-') {
+                  fprintf(stderr, "Error: PID is negative \"%s\".\n", pid);
```
hishamhm: The user wouldn't be able to see these messages since they would be quickly erased by the full screen interface. We can't expect users to pipe stderr just to catch errors.

hishamhm:
Hi!

Thank you for taking the time and sending a patch! I posted some code-review comments above, but my main question is: is the additional logic really necessary? htop tries to produce a pleasant interface, but it also tries to remain a small tool in memory footprint, so every little addition has to be considered in terms of the value it brings. From what I can see, the situation for users entering invalid data wouldn't change a lot under this patch.

As the code currently is, any invalid values entered are silently ignored. If we really want to change this behavior, the approach I suggest would be to just replace `atoi()` with `strtoul()`, add one `if` to check for any problems (i.e. to see if it parsed up until the next ',' or the end of string), and then exit with "invalid argument for -p" in case of errors of any kind.

htop.c
GiedriusS: strtoul() returns `unsigned long`. `unsigned long` could be bigger than `unsigned int`. It could be produced when a user enters a number bigger than UINT_MAX.

htop.c
GiedriusS: These fprintf's are in spirit of the other option handling when in the case of malformed input a message is printed to stderr.

GiedriusS:
On Fri, Feb 19, 2016 at 03:50:24PM -0800, Hisham Muhammad wrote:
> Hi!
> 
> Thank you for taking the time and sending a patch! I posted some code-review comments above, but my main question is: is the additional logic really necessary? htop tries to produce a pleasant interface, but it also tries to remain a small tool in memory footprint, so every little addition has to be considered in terms of the value it brings. From what I can see, the situation for users entering invalid data wouldn't change a lot under this patch.

The only difference in terms of memory is that there is an extra `char*` and one
`unsigned int` was converted to `unsigned long`. I don't think that is much :)
Well I think good programs should tell the user when something goes wrong so
currently when malformed input is given to htop via "-p" it could: a) invoke
undefined behaviour; b) be silently truncated to a smaller value. It would be
nice to tell the user when the input is wrong and gracefully deal with it
instead of just accepting it.

>
> As the code currently is, any invalid values entered are silently ignored. If we really want to change this behavior, the approach I suggest would be to just replace `atoi()` with `strtoul()`, add one `if` to check for any problems (i.e. to see if it parsed up until the next ',' or the end of string), and then exit with "invalid argument for -p" in case of errors of any kind.
How are they ignored? It seems to me that any kind of value atoi() produces is
passed to the other upper layers.

I need to remind you that if atoi() fails then the behaviour is undefined and
bad things may happen. From a C99 draft I have:
"The functions atof, atoi, atol, and atoll need not affect the value of the
integer expression errno on an error. If the value of the result cannot be
represented, the behavior is undefined."
Also in "J.2 Undefined behaviour" (a example of undefined behaviour):
"-- The value of the result of converting a string to a number by the atof,
atoi, atol, or atoll function cannot be represented"

With this patch we also check if the number given by the user isn't too big and
won't be truncated which I'm sure any user wants to avoid.

>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/hishamhm/htop/pull/418#issuecomment-186457316

-- 
        Giedrius

Explorer09:
May I comment on this thing?
Yes I know you are trying to address a problem about number overflow or invalid number in `-p` argument. But I think we can simplify the behavior a little bit:
* Making this error fatal and exit immediately. There is no sense for htop to continue when user specifies an overflow/out-of-range argument. Such case is mostly intentional or due to shell scripting error. (This is unlike the case where PIDs in argument are valid but not used.)
* Just emit one form of error message. It's not useful to distinguish "PID is negative" and "PID out of range" and "PID argument containing invalid characters". All of these can be simply 'Invalid PID argument(s) "%s".' so we can save some bytes having different error messages in binary.

GiedriusS:
On Sun, Feb 21, 2016 at 04:20:24AM -0800, Kang-Che Sung (宋岡哲) wrote:
> May I comment on this thing?
> Yes I know you are trying to address a problem about number overflow or invalid number in `-p` argument. But I think we can simplify the behavior a little bit:
> * Making this error fatal and exit immediately. There is no sense for htop to continue when user specifies an overflow/out-of-range argument. Such case is mostly intentional or due to shell scripting error. (This is unlike the case where PIDs in argument are valid but not used.)
Not sure about this. In other cases of that switch we don't exit if there is
errorneous input. But this would make htop more in line with what top does like
you have described: exit if a specified PID is invalid. So this is a
possibility. It all depends on what @hishamhm thinks.

> * Just emit one form of error message. It's not useful to distinguish "PID is negative" and "PID out of range" and "PID argument containing invalid characters". All of these can be simply 'Invalid PID argument(s) "%s".' so we can save some bytes having different error messages in binary.
It seems like you and @hishamhm agree on making the error messages simpler. I
guess this is the route we could go down.

hishamhm:
> Not sure about this. In other cases of that switch we don't exit if there is errorneous input. But this  would make htop more in line with what top does like you have described: exit if a specified PID is invalid. So this is a possibility. It all depends on what @hishamhm thinks.

What top does is not a concern (there are several tops around anyway, which one? ;) ). But yes, just failing on invalid entries is okay and would work better.

htop.c
hishamhm: Argh, I need to stop thinking in 32-bits!

GiedriusS:
Pushed a updated version. Put all checking inside one function; exits with 1 and prints only one message in case of errorneous input.

htop.c
```diff
@@ -53,6 +55,25 @@ static void printHelpFlag() {
    exit(0);
 }
 
+static bool parsePID(char *pid_str, unsigned int *result) {
+   char *endptr;
+   unsigned long tmp_result;
+
+   if (pid_str[0] == '-')
+      return false;
```
hishamhm: won't strtoul below catch this case?

htop.c
```diff
@@ -53,6 +55,25 @@ static void printHelpFlag() {
...
+
+   errno = 0;
+   tmp_result = strtoul(pid_str, &endptr, 10);
+   if (*endptr != '\0' || errno != 0)
+      return false;
+
+   if (tmp_result > UINT_MAX)
+      return false;
```
hishamhm: this test is also unnecessary, because this doesn't really check for the range of valid PIDs anyway. Given any valid number can be a non-existent PID, I don't think it's necessary or useful to validate PID by its numeric value.

As we suggested before, it's better to simplify all these tests, not just move them all over to a separate function.

htop.c
Explorer09: No it won't. `strtoul` returns an absolute value (i.e. ignores the sign) of the integer.

Explorer09: No it won't. `strtoul` accepts negative numbers but will be cast to unsigned. (Similar to `(unsigned long)stroul(str,endptr,base)` )

Explorer09:
Would you please try my version?
Here: https://github.com/Explorer09/htop/tree/opt-p-error

htop.c
Explorer09: The check of UINT_MAX isn't to make sure whether the PID is valid. In my code (see post below), I made this check also. It's to ensure that the number can be safely type-demoted to 'unsigned int' and be put in the Hashtable.

Determining the maximum PID number is platform specific, as far as I know. For Linux it's `/proc/sys/kernel/pid_max` and PID_MAX_LIMIT. But for me, I don't wan't to include a platform specific header or read a file just to check this damn limit.

http://stackoverflow.com/questions/6294133/maximum-pid-in-linux
http://unix.stackexchange.com/questions/16883/what-is-the-maximum-value-of-the-pid-of-a-process

Still, I don't like @GiedriusS 's use of `parsePID` function. It adds another "interface" that is not necessary.

GiedriusS:
I still don't why you don't want to have simple function for this like I
have made. It would also come in handy to replace other atoi() calls later
so we wouldn't have to repeat the code.

Explorer09:
@GiedriusS 
Because I'm not ready to make it a utility function with such a clear interface. In this use case our error is fatal, but in other cases we may ignore errors or not need to check errors at all.

YAGNI (You aren't gonna need it) principle. If we find a use case that require us to turn that into a function, we will do it then.

GiedriusS:
Yes, it's simple and clear and it does what it says on the tin. I think you
are a bit confused because the function doesn't exit() itself, the caller
does that. It is their own responsibility to do the "right" thing according
to the output of parsePID(). I already told you one use case - we need to
stop using atoi() in configuration parsing as well.

Explorer09:
@GiedriusS 
If we are to make a function for that, it would be a generic `strtoi`, and avoid your (unextendable) boolean return value and bring the `errno` back, which is better designed.

PR #420, opened 2016-02-21
regex support for filtering and search

mscalora:
I've implemented regex support for filtering and search and also provide a way to exclude, like a negative filter. By default, filters are treated as a regex. If the first character is a `$`, the rest is treated as a plain, literal string. A prefix of [!] makes the regex exclude processes (negative filter). Lastly `$!` would create a negative, plain string filter.

Examples:

`bash|zsh`

`\.jpe?g`

`$.gif`

`!/sbin/|/libexec/`

`$!.pid`

Note: I thought about adding an option for regex but I don't think there is a conflict very often. I think the only meta-char that comes up very often is `.` and it being a wildcard char, you usually find what you want anyways.

Let me know that you think, I can do some rework it it is needed and I would be happy to update the help/docs once I know you are open to this feature. I built and tested on Ubuntu and Mac OSX.


GiedriusS:
Something is wrong with your PR. For some reason there are a bunch of commits from the master branch in your PR.

GiedriusS:
Maybe update the manual page as well?

GiedriusS:
Because '$' and '!' are part of regular expression syntax, maybe there could be two hotkeys when in "search mode" for this funcionality:
F4 -> invert search
When you push it, the tooltip text changes to inform you that the search is now inverted:
F4 -> regular search

F5 -> simple string
F5 -> regexp

hishamhm:
Hi @mscalora! This is an interesting feature, and I think it could make a lot of people happy.

About the implementation, maybe you could avoid the `Filter_new`/`Filter_delete` in the ProcessList by storing it inside the IncSet (replacing the `const char* incFilter` field with a `Filter* filter` field). If this is doable, then maybe Filter could go inside IncSet altogether.

As for the UI, I think the suggestions of @GiedriusS are great (I'd just make F6 the key for invert/regular to avoid reusing the keys F3 and F4, which enter these modes). Having these modes integrated in the UI are much more in the htop spirit than creating custom syntax as special prefixes.

One last piece of feedback: your diff is showing against the 2.0.0 tag, not the latest `master`. The diff is really big, showing 21 changed files — can this be adjusted to be a diff against the latest `master`?

mscalora:
I reimplemented based on feedback, see https://github.com/hishamhm/htop/pull/428

hishamhm:
Supplanted by #428.

PR #427, opened 2016-02-27
Fix low CPU usage reporting in FreeBSD

### LINKED ISSUE #400: FreeBSD CPU usage too low
Output from `top -PI`
```
last pid: 80618;  load averages:  2.24,  2.66,  1.99                                                           up 28+05:04:54  22:09:47
120 processes: 3 running, 117 sleeping
CPU 0: 23.6% user,  0.0% nice, 17.7% system,  0.0% interrupt, 58.7% idle
CPU 1: 48.8% user,  0.0% nice, 17.7% system,  0.0% interrupt, 33.5% idle
CPU 2: 45.7% user,  0.0% nice, 14.6% system,  0.0% interrupt, 39.8% idle
CPU 3: 30.7% user,  0.0% nice, 17.3% system,  0.0% interrupt, 52.0% idle
```
Output in htop
```
  1  [||||||||                                             12.1%]   Tasks: 516, 0 thr; 7 running
  2  [|||||||                                              10.0%]   Load average: 2.26 2.67 1.99
  3  [|||||                                                 7.0%]   Uptime: 28 days, 05:04:47
  4  [|||||                                                 7.5%]
```
This is during a Poudriere bulk run (2 make processes parallel, each with make -j2) on a system with 4 CPUs (2+HT)

My hunch is that this is divided by the number of CPUs.

tcreech:
Scope issue resulted in per-CPU click totals being accumulated across all CPUs, which meant CPU usages were divided by increasing amounts. I believe this patch fixes #400.

### DIFF
--- a/freebsd/FreeBSDProcessList.c
+++ b/freebsd/FreeBSDProcessList.c
@@ -212,9 +212,6 @@ static inline void FreeBSDProcessList_scanCPUTime(ProcessList* pl) {
    unsigned long     *cp_time_n; // old clicks state
    unsigned long     *cp_time_o; // current clicks state
 
-   unsigned long long total_o = 0;
-   unsigned long long total_n = 0;
-   unsigned long long total_d = 0;
    unsigned long cp_time_d[CPUSTATES];
    double        cp_time_p[CPUSTATES];
 
@@ -251,6 +248,9 @@ static inline void FreeBSDProcessList_scanCPUTime(ProcessList* pl) {
       }
 
       // diff old vs new
+      unsigned long long total_o = 0;
+      unsigned long long total_n = 0;
+      unsigned long long total_d = 0;
       for (int s = 0; s < CPUSTATES; s++) {
         cp_time_d[s] = cp_time_n[s] - cp_time_o[s];
         total_o += cp_time_o[s];


hishamhm:
Thank you!! Merged! :D

myfreeweb:
This fixes the overall system CPU usage reporting, which is nice :) But per-process reporting is still wrong:

screen_2016-02-28-20 11 51

(on this screenshot, erlang is definitely the thing that puts the system under full load, but it's 0.0% in the CPU column)

I might just try fixing it…

tcreech:
My guess is that this is a separate problem seen when multi-threading is involved, and that htop is only reporting the CPU usage for the first thread. @myfreeweb, on your system from the screenshot, is it possible that the first thread is pretty idle, and child threads are very busy?

(E.g., what's the output of `ps -H -o pid,nlwp,lwp,comm,pcpu -p 801` about 20 seconds after the erlang process has been running?)

myfreeweb:
Yeah, the first thread is pretty idle, but I don't think that's the problem.

Here's what htop does currently:
```c
      proc->percent_cpu = 100.0 * ((double)kproc->ki_pctcpu / (double)kernelFScale);
```

ps does this:
```c
#define	fxtofl(fixpt)	((double)(fixpt) / fscale)

	/* XXX - I don't like this */
	if (k->ki_p->ki_swtime == 0 || (k->ki_p->ki_flag & P_INMEM) == 0)
		return (0.0);
	if (rawcpu)
		return (100.0 * fxtofl(k->ki_p->ki_pctcpu));
	return (100.0 * fxtofl(k->ki_p->ki_pctcpu) /
		(1.0 - exp(k->ki_p->ki_swtime * log(fxtofl(ccpu)))));

```

So, htop does "raw cpu". If I change the htop formula to the non-raw variant (hardcoded cpu count to 4 for testing):
```c
      proc->percent_cpu = 100.0 * ((double)kproc->ki_pctcpu / (double)kernelFScale) /
         (1.0 - exp(kproc->ki_swtime * log((double)4 / (double)kernelFScale)));
```

It's better. I think it's calculated properly in that case.

But over very short intervals!! I see 99% usage for a moment on programs that are launching. And a tight math loop in a python repl slowly ramps up to 25% usage. Which is weird too, it should show 25% instantly.

FreeBSD's top actually saves the previous kproc and calculates the difference.

tcreech:
Here's what I think is happening: in FreeBSD, the thread ID space is not shared with the process ID space, unlike on Linux. As a result, when htop calls `kvm_getprocs` using `KERN_PROC_ALL`, all individual threads are reported but some threads share the same PID. When htop slurps in everything yielded by `kvm_getprocs`, it erroneously concludes that all threads in a process are duplicates because they share a PID. The result is that only a single thread and its CPU consumption are reported.

A quick fix is to use `kvm_getprocs` with `KERN_PROC_PROC` rather than `KERN_PROC_ALL`: this will report only one item per PID, resulting potentially in `ki_pctcpu` > 100. (E.g., https://github.com/tcreech/htop/commit/cc8375f .) With this, CPU usage reporting should be accurate, but threads within processes are not visible to htop.

This doesn't affect the "ramping up" behavior, which I believe is yet another separate issue.

PR #428, opened 2016-02-28
Added regular expression functionality to filters with function key to toggle on/off, added function key to toggle include/exclude for filters. Updated man page for above and fixed a few typos and grammar issues.

mscalora:
Added regular expression functionality to filters with function key to toggle on/off, added function key to toggle include/exclude for filters. Updated man page for above and fixed a few typos and grammar issues.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -225,9 +225,8 @@ static Htop_Reaction actionToggleTreeView(State* st) {
 }
 
 static Htop_Reaction actionIncFilter(State* st) {
-   IncSet* inc = ((MainPanel*)st->panel)->inc;
-   IncSet_activate(inc, INC_FILTER, st->panel);
-   st->pl->incFilter = IncSet_filter(inc);
+   st->pl->incSet = ((MainPanel*)st->panel)->inc;
+   IncSet_activate(st->pl->incSet, INC_FILTER, st->panel);
    return HTOP_REFRESH | HTOP_KEEP_FOLLOWING;
 }
 
--- a/IncSet.c
+++ b/IncSet.c
@@ -17,6 +17,7 @@ in the source distribution for its full text.
 
 #include "FunctionBar.h"
 #include "Panel.h"
+#include <regex.h>
 #include <stdbool.h>
 
 #define INCMODE_MAX 40
@@ -40,6 +41,11 @@ typedef struct IncSet_ {
    IncMode* active;
    FunctionBar* defaultBar;
    bool filtering;
+   bool regex;
+   bool exclude;
+   bool regexAlloc;
+   char* rawFilter;
+   regex_t regexFilter;
 } IncSet;
 
 typedef const char* (*IncMode_GetPanelValue)(Panel*, int);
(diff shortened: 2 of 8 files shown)


Explorer09:
1. Could you please bring the typo fixes in man pages out and make a separate commit with them? That way it can be cherry-picked into an htop bug fix release. Also it makes the commit less confusing (when reading the code changes).
2. For the UI, I'm personally confused about the behavior when the regex and the inversion are _combined_. Although it's a feature, are users going to use that frequently? And with that implemented, we need a reason why case insensitive matching is not there. I wish to avoid feature bloat (replicating too many things that `ps | grep` can do) and add features only if there are frequent use cases with them.

mscalora:
1. I'll do that

2a. The inversion is very simple, normally matches are shown and others hidden, when inverted matches are hidden and others show. It applies the same way for regex and plain strings. My frequent usage is to hide a lot of nuisance processes with something like `^/App|^/Sys`. For me, plain strings are of must less utility, the only RE metachar that comes up in process/commands very often is period and I can always escape it like `\.app` or `com\.apple`
 
2b. The old filters (and search were case insensitive), I don't think there's a lot of value in any case sensitive filtering so it is all case insensitive.

try googling  `htop regex` , I'm confident there's many users who want **regex** and **filter out**

Explorer09:
@mscalora 
For the second point, I still cannot rule out the possibility of expanding the filter to something more advanced. For your info, `top` already has an advanced filter. It just doesn't support regex yet.

`top`'s filter can work with any field, stack multiple filters together, match case-sensitively or insensitively, and allows both inclusion and exclusion.
It's filter syntax is this (see [man page](http://man7.org/linux/man-pages/man1/top.1.html) if you wish to know more): press 'o' or 'O' to add a filter (lowercase indicates case-insensitive filter), then type this

    [!]<field name><op><value>
    <op> could be '=', '<', or '>'. '=' allows partial matches.
    '<' and '>' compare in lexicographical order and so can work with strings 
    ! indicates 'exclusion'
    Example: 'VIRT>10000' + '!USER=root'

Ctrl-O lists the current filters, '=' clears all filters. This the filter system in `top`.

mscalora:
As someone who bounces between Mac and Linux, **top** is one of those painful commands that is very different. I find the gnu version powerful but I have to reread the man page every time I want to use it, the Mac version is much more simple and doesn't meet my needs.

hishamhm:
@mscalora The patch looks much nicer now, with all the regex support self-contained into `IncSet` and the function keys controlling the behavior! Thank you!

As for case-sensitivity, using case-insensitive searches has worked pretty well for us and I think it can stay this way. 

As for the specifics of the UI, I need to give it a spin to see how it feels (and maybe tweak a label or two) but from what I'm reading in the source I think everything's in the right place!

@Explorer09 I think `top`'s filtering system is way over-engineered. If anything, I look at `top` for inspirations of what _not_ to do in terms of design philosophy.

mscalora:
Really wishing this would get merged in!

NorseGaud:
This this still a possibility?

rwlaschin:
It would be great to merge this in.

nrvale0:
Is this just waiting on merge conflict resolution?

PR #430, opened 2016-03-02
introduce Battery Meter_display() and color warning and critical level

eworm-de:


Explorer09:
First, I do _love_ the idea of having different color for battery charging status, discharging status, warning and critical level. But I'm not convinced that we need two levels of warning and that warnings are based on percentages (for your code, 10% and 5%). I want to know how GUIs and desktop environments do their battery warnings. Do they display warnings on percentages, or on estimated remaining time?

eworm-de:
Running Xfce here which works with percentage based levels.

Remaining uptime depends on the workload. So changing workloads can decrease and *increase* remaining time. This way thresholds could be crossed over and over again. (`htop` changing the color more than once would not be that disaster. Notifications from desktop environment can be really annoying, though.) And calculating the time would be a lot more complex.

Explorer09:
@eworm-de Thanks for info with Xfce. However I don't think calculating the time would be something complex if we would reference other implementations. The little [`acpi` command](https://sourceforge.net/projects/acpiclient/) does provide remaining time for us, for example.
Anyway, the remaining time feature is secondary for htop.

By the way, [this](http://git.xfce.org/panel-plugins/xfce4-battery-plugin/tree/panel-plugin/battery.c) should be source code of the Xfce battery display you were talking about. I see the percentage are tweakable there, but default to 5% and 10% as you say.

Explorer09:
Another comment here. @eworm-de

Your current battery meter code only changes the color when in text display mode. Will it be better if we have it in bar mode and graph mode too?

I have an idea of how it can be done. But it will increase `maxItems` of battery meter to 3, and `BatteryMeter_attributes[] = {BATTERY, BATTERY_WARN, BATTERY_CRIT}`.

hishamhm:
> Anyway, the remaining time feature is secondary for htop.

?

I never got around to implementing the display/calculation of remaining-time estimates, but in fact I consider it a lot more useful than color-coding percentages (which will always be somewhat arbitrary).

> I have an idea of how it can be done. But it will increase maxItems of battery meter to 3

I see what you have in mind, but I don't think the cosmetic tweak is worth the complication. A nearly-empty bar is visually clear enough.

PR #435, opened 2016-03-06
Improve error reporting on FreeBSD libkvm call

mmcco:
This involves switching from kvm_open(3) to kvm_openfiles(3). The only
difference is that the latter has saner error reporting (see the man
page for details). We can now fatally report the error rather than just
calling assert(3).


mmcco:
I can't actually test this at the moment because I don't run FreeBSD. Someone should before it gets merged.

hishamhm:
Tested in PC-BSD 10. Thanks for the patch!

PR #436, opened 2016-03-06
Misc. OpenBSD tuneup and improvement

mmcco:
Including:

 o set *basenameEnd even in error cases (FreeBSD probably needs this)

 o use kvm_openfiles(3) rather than kvm_open(3) so that we can report
   errors (as with FreeBSD)

 o sanify the process argument list creation by using strlcat(3)

 o drop the pageSizeKb variable and use the PAGE_SIZE_KB macro directly,
   as the page size can't change anyway

 o clean up a few macros, add MINIMUM() and MAXIMUM() (should be
   mirrored to FreeBSD)

 o fix some syntax

 o add some useful comments

### DIFF
--- a/openbsd/OpenBSDProcessList.c
+++ b/openbsd/OpenBSDProcessList.c
@@ -12,14 +12,15 @@ in the source distribution for its full text.
 
 #include <unistd.h>
 #include <stdlib.h>
-#include <err.h>
-#include <errno.h>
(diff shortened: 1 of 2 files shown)


hishamhm:
Thank you! Merged! Making the final run through PRs to get 2.0.1 out.

PR #437, opened 2016-03-06
De-uglify copyright notices of OpenBSD files

mmcco:


hishamhm:
Thanks for the suggestion, but I prefer to keep it in other style, consistent with the rest of the codebase.

mmcco:
Would you consider accepting a patch that fixes all of them?

Issue #438, opened 2016-03-07
Side effect: Alt+O, Alt+P as PageDown and PageUp


Explorer09:
Commit c23d4f12d , which defines Alt key combinations as (KEY_F(60)+(c-'A')) actually overflows with my ncurses version (my <curses.h> says that KEY_F(63) is the maximum). This brings a funny side effect that the key codes of Alt+O and Alt+P collide with PageDown and PageUp, respectively.

This is not necessary a bug, but it's better to have this side effect documented or avoid it.

hishamhm:
Good catch, thanks

Explorer09:
f295a52 fixes nothing. You just made them Alt+K and Alt+L instead, which is worse as they collide with Alt+{H,J,K,L} scrolling.
The right way would be: `#define KEY_ALT(x) KEY_F(64-26) + (x - 'A')`

PR #441, opened 2016-03-09
Avoid overlapping key values defined by curses (Real fix).

Explorer09:
Real fix for issue #438.


hishamhm:
Oops, got it now! Thanks for re-catching it. ^_^

PR #442, opened 2016-03-09
Replace platform-specific PID_MAX code with sizeof(pid_t)

mmcco:
It seems easier to use sizeof(pid_t) instead of platform-specific code
with magic numbers. PID_MAX generally isn't exposed to userland by
sys/proc.h, so porters have to manually copy the number over and are
unlikely to update if it changes.

It seems that it costs us a few characters worth of screen space,
though. Hisham will decide whether it's worth the simplification. I
highly doubt that any platform has pid_t larger than 32 bits, so this is
unlikely to waste more than ~5 characters.

For what it's worth, POSIX tells us that pid_t is:

 o signed
 o no larger than long

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -158,7 +158,7 @@ void Process_delete(Object* cast);
 bool Process_isThread(Process* this);
 extern ProcessFieldData Process_fields[];
 extern ProcessPidColumn Process_pidColumns[];
-extern char Process_pidFormat[20];
+extern char Process_pidFormat[21];
 
 typedef Process*(*Process_New)(struct Settings_*);
 typedef void (*Process_WriteField)(Process*, RichString*, ProcessField);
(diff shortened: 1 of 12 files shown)


mmcco:
Also, I think my size increase of `Process_pidFormat` is valid, but whoever wrote that code (probably Hisham) will know far better than I.

Explorer09:
Just for the info. In Linux there are two limits on the maximum PID. First one is `/proc/sys/kernel/pid_max` which is configurable by system administrator. Second one is [PID_MAX_LIMIT](http://lxr.free-electrons.com/ident?i=PID_MAX_LIMIT), the hard-coded absolute limit in the kernel.

I don't think using sizeof(pid_t) will do what we want better.

See also these 2 StackOverflow questions:
http://stackoverflow.com/questions/6294133/maximum-pid-in-linux
http://unix.stackexchange.com/questions/16883/what-is-the-maximum-value-of-the-pid-of-a-process

mmcco:
I think you're missing my point: what defines better? Is it worthwhile to have a bunch of platform-specific code that sets more accurate limits? We know that a PID can't be larger than the limit of the int type it's stored in (`pid_t`), so as long as we don't actually need more accuracy, `sizeof(pid_t)` works fine.

Explorer09:
@mmcco What you did extends the PID column width unnecessarily. In my htop, 5 characters are already enough. But what you did will make the column 11 characters wide. Do I analyze anything wrong with your code?

mmcco:
Nope, you're right. That's exactly what I described in the commit message above.

Explorer09:
And one more thing I'm going to be picky: `sizeof(pid_t)` is a compile-time constant. So your `default: errx(1, "sizeof(pid_t) is bizarre - neither 4 nor 8");` line will probably never be reached. However your error line looks more like a user error than an assertion or compile-time error.

hishamhm:
11 (or 20!) characters for PIDs in platforms that don't need this much is way too wide. While I appreciate the amount of code this removes and having one less platform-specific thing to worry about, the screen real estate we save by being more specific is worth the price. So sorry, but I'll close this one.

mmcco:
Makes sense. I was conflicted too. Thanks for reviewing.

PR #443, opened 2016-03-09
Add Solarized color scheme

ngg:
This PR adds a new color scheme that looks good in Solarized terminals.
It is a slight modification of the Black Night scheme.
Based on https://bitbucket.org/ajenkins/htop-solarized/src/


willvousden:
This seems to be a sensible addition.  The current "Broken Gray" theme seems to be a compromise, since it doesn't make full use of the Solarized palette.

Solarized is an excellent colour scheme, but due to xterm limitations it requires applications to use a specifically designed theme.  Since Solarized is so widely used, it seems to make sense to include such a theme in htop (since user-defined themes aren't supported).

antoineco:
The end result looks like this on macOS' Terminal.app with a Solarized Dark color scheme:
(**bold** text is displayed with *bright* colors, just like the GNOME Terminal)

htop_solarized_pr443

I find the usage of cyan excessive, but at least all the text is readable.

antoineco:
In comparison, this is how the scheme renders with the following changes, in the same conditions as above:

```diff
--- a/CRT.c
+++ b/CRT.c

   [COLORSCHEME_SOLARIZED] = {
-      [RESET_COLOR] = ColorPair(Cyan,Black),
+      [RESET_COLOR] = ColorPair(White,Black),
-      [DEFAULT_COLOR] = ColorPair(Cyan,Black),
+      [DEFAULT_COLOR] = ColorPair(White,Black),
```

htop_solarized_pr443_white_text

willvousden:
Can we expect this PR to be merged?

PR #444, opened 2016-03-11
Explicit "maxItems" property of meters

Explorer09:
Two commits included in this pull request:
Fist commit (7d3f67e) reverts 5c593fa (xCalloc) because the second commit _will_ expect a `nmemb` argument to calloc be zero.
Second commit (99fb307) is the main one. Make "maxItems" property of meters explicit, and allows its value be zero for meters that have no numeric values at all (e.g. HostnameMeter, BlankMeter).

Pull request #355 (graph-color-new) will be based on this pull request.


hishamhm:
Maybe xCalloc should fail if `nmemb > 0 && size == 0`? The text in the calloc manpage is a bit ambiguous, so I tried to make xCalloc stricter to be on the safe side:

```
If nmemb or size is 0, then calloc() returns
either NULL, or a unique pointer value that can later  be  successfully
passed to free().
```

I can read this two ways:
* `calloc(0, 1)` should return NULL and `calloc(1, 0)` should return a pointer
* both of them may return NULL or a pointer and we never know

The first reading makes more sense and so `calloc(0, 1) == NULL` should be allowed but `calloc(1, 0) == NULL` should abort.

hishamhm:
On further testing, it seems that both of them returned pointers!

```c
#include <stdio.h>
#include <stdlib.h>
main() {
   printf("calloc(0,1) == %p\n", calloc(0, 1));
   printf("calloc(1,0) == %p\n", calloc(1, 0));
}
```

```bash
hisham@pointer ~/junkcode]gcc -o calloc calloc.c
hisham@pointer ~/junkcode]./calloc              
calloc(0,1) == 0x92b8008
calloc(1,0) == 0x92b8018
```

Explorer09:
@hishamhm
Your second interpretation "both of them may return NULL or a pointer and we never know" is the right one. :)

Whether it returns NULL or a pointer depends on library implementation, but as far as we know, the NULL return value in such case are _not_ errors.

See how POSIX says about calloc:

> Upon successful completion with both nelem and elsize non-zero, calloc() shall return a pointer to the allocated space. If either nelem or elsize is 0, then either a null pointer or a unique pointer value that can be successfully passed to free() shall be returned. Otherwise, it shall return a null pointer and set errno to indicate the error. 

http://pubs.opengroup.org/onlinepubs/009695399/functions/calloc.html

hishamhm:
Ah, well, C libraries... I guess we can count on them for having weird specifications.

Given that a 0-sized pointer can't really be used for anything (dereferenced) and `free(NULL)` is harmless, I guess ignoring whether it returned NULL or a value and simply freeing the result is fine for the case you're using there.

Issue #434, opened 2016-03-05
Kernel version as htop meter


Ape:
I'd like to have the kernel version available as one of the meters in htop. The output should be something similar to `uname -sr`:

> Kernel: **Linux 4.4.1-2-ARCH**

jjungo:
@hishamhm About colors, I don't know if this is helpful in this case.

If my patch isn't so bad, I can go ahead with others platforms, but I can't test it.

PR #448, opened 2016-03-15
Kernel version meter

jjungo:
Small fix for issue #434.
Only for Linux


Explorer09:
I'm confused about your implementation. You seem to grab the 1st and 3rd fields from the Linux-specific `/proc/version` file. But isn't that more portable to just use the [uname(2)](http://linux.die.net/man/2/uname) system call to get the same thing? [It's POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/functions/uname.html), so you may expect it to work in many platforms. And if you are not sure what it will output on various systems, [check the Wikipedia article for examples](https://en.wikipedia.org/wiki/Uname). Hope that helps.

jjungo:
I didn't know about this syscall. Your implementation #491 seems to be better than mine, I'll close this PR. Thank you for the tips.

PR #452, opened 2016-03-19
Assert (Platform_meterTypes[0] == &CPUMeter_class)

Explorer09:
Just assume Platform_meterTypes[0] is always &CPUMeter_class for every
platform. This removes a conditional in AvailableMetersPanel_new().

Also adds some comments about the logic here. Without assuming
Platform_meterTypes[0], the (int i=1) clause in this for loop will not
make sense. (I.e. Why not (int i=0)? )

Also replaced a sprintf() call with safer snprintf() in code further
below.


hishamhm:
A most reasonable tweak, thanks :)

PR #463, opened 2016-04-07
Assert (Meter.total > 0)

Explorer09:
Adds assertions in BarMeterMode_draw() and GraphMeterMode_draw()
These two would seem trivial, but can help detect divide-by-zero and
bugs in meters code.
Also meant to convey a message: That we are not supporting dynamical
maximum in bars or graphs yet.


Explorer09:
Closing this one because of pull request #487

PR #464, opened 2016-04-07
Add support of UUID CTID used in the new version of OpenVZ

TresRus:
In the new version of OpenVZ (Virtuozzo 7), containers could be identified by UUIDs and integer numbers like before (so called legacy CTID). Existing support of OpenVZ can't work with UUIDs. This patch fix it.

CTIDs and VPIDs are now read from /proc/[id]/status instead of /proc/[id]/stat because of the following reasons:
- the /proc/[id]/status has named fields and OpenVZ kernel writes either CTID or UUID to envID field (depending on which type is used);
- OpenVZ kernel appends own data to /proc/[id]/stat file, a change in parameters reported by vanilla kernel would also shift the position of OpenVZ data so it would not be read correctly.

As UUIDs are much longer than old numerical CTIDs, the UUID strings will get truncated in output (although sorting is still done using complete UUIDs).

### DIFF
--- a/linux/LinuxProcess.c
+++ b/linux/LinuxProcess.c
@@ -114,7 +114,7 @@ typedef struct LinuxProcess_ {
    double io_rate_write_bps;
    #endif
    #ifdef HAVE_OPENVZ
-   unsigned int ctid;
+   char* ctid;
    unsigned int vpid;
    #endif
    #ifdef HAVE_VSERVER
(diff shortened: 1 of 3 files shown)


fasterit:
This patch (which we carry for Debian) collides with the new thread detection code now, cf. commit 47cf1532b0c9fbc70bada5022a7db07d3cc4811a (bug #761).

So

1. This patch needs updating and
2. Q to Hisham: Is this functionality interesting for htop to include or will it not make it in tree? (as to whether we need to / should continue to carry it in Debian)

/DLange

PR #474, opened 2016-04-19
Upstream/desktop keywords

fasterit:


fasterit:
two commits sneaked in there, trying again

PR #476, opened 2016-04-19
Use SOURCE_DATE_EPOCH instead of current year in copyright message to…

fasterit:
… make build reproducible

Debian patch from
https://anonscm.debian.org/cgit/collab-maint/htop.git/tree/debian/patches/003-use-source-date-epoch.patch
by Graham Inggs

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4,7 +4,7 @@
 AC_PREREQ(2.65)
 AC_INIT([htop],[2.0.1],[hisham@gobolinux.org])
 
-year=$(date +%Y)
+year=$(date --utc --date="@${SOURCE_DATE_EPOCH:-$(date +%s)}" +%Y)
 
 # The following two lines are required by hwloc scripts
 AC_USE_SYSTEM_EXTENSIONS


fasterit:
Alexis Bienvenüe has proposed a similar patch in [Debian bug #822963](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=822963) that may be more robust on BSDs. Could be worth checking on these platforms. Feedback welcome here and/or on the Debian bug report.
The essence is:
```
+-year=$(date +%Y)
++SOURCE_DATE_EPOCH="${SOURCE_DATE_EPOCH:-$(date +%s)}"
++year=$(date -u -d "@$SOURCE_DATE_EPOCH" "+%Y" 2>/dev/null || date -u -r "$SOURCE_DATE_EPOCH" "+%Y" 2>/dev/null || date -u "+%Y")
```
... to make htop 2.0.1+ build reproducibly.
Also see https://tests.reproducible-builds.org/rb-pkg/unstable/amd64/htop.html while builds are still unreproducible.

hishamhm:
@fasterit Thank you! I'll use these two lines from your last comment.

PR #487, opened 2016-05-04
Graph meter coloring & dynamic scaling!

Explorer09:
**This branch needs rebasing and it's code needs some polishing.**

This pull request is much bigger than #355 because it's improved and supports dynamic scaling. :)

Screenshot:
Screenshot

Meters that will use what I called "dynamic scaling":
- Load meter & Load average meter
- Tasks meter (a.k.a. Task counter)
- Uptime meter

Caveats:
- When merging, it's better to review changes in every commit as some of them changes variable naming in existing code for some reasons (that you might disagree; please tell me).
- The colors array in GraphData will take larger size than necessary. Although I have optimized the (tree) data structure so that it allocated no more "cells" than needed, the cell size is still large and not optimized yet. The color cells uses `ColorElements` enum type (which is sizeof(int) by default), and I didn't yet choose a way to optimize it. I'm asking suggestions about what to do here in code.*

*: There are two way of optimization in code:
(1) changing ColorElements typedef from `enum` to `short` or `char`, or
(2) use `__attribute__((packed))` compiler directive on the enum (which will require compiler to support GCC extension), or
(3) just don't bother at all.


hishamhm:
Oh wow, this will take me a while to go through! But I'm impressed :)

eworm-de:
Any chance we see this being merged any time soon?
I really like the graph meter coloring but I do not like to resolve merge conflicts again and again (or wait for @Explorer09 to do it).

Explorer09:
@eworm-de 
Hi, I was not maintaining this branch for a while, partly because @hishamhm [was not interested in merging](https://github.com/hishamhm/htop/pull/651#issuecomment-318154184), and because I have been troubled about variable naming for a while.
I would appreciate if someone likes to help making this PR branch up to date.

Issue #471, opened 2016-04-18
Proposal to rename *Meter_setValues() functions to *Meter_updateValues()


Explorer09:
The function name "setValues" is misleading. For most OOP (object-oriented programming) contexts, setXXX functions mean they will change some member variables of an object into something specified in function arguments. But in the *Meter_setValues() case, the new values are not from the arguments, but from a hard-coded source. The caller is not supposed to change the `values[]` to anything it likes, but rather to "update" the values from the source. Hence, `updateValues` is a better name for this family of functions.

If permitted, I can file a pull request to change the names of these.

hishamhm:
Yes, I agree. This is a sensible change.

I haven't slotted a lot of time for htop lately (I usually do this is "batch runs" rather than constant bit-by-bit maintenance, and the 2.0.0/2.0.1 cycle took me a _lot_ of effort), but I'll run through the more delicate PRs when I do my next batch run. This one, however, seems an easy change to apply.

PR #488, opened 2016-05-04
Rename Meter.setValues() functions to updateValues()

Explorer09:
Rationale (copied from htop issue #471):
The function name "setValues" is misleading. For most OOP (object-
oriented programming) contexts, setXXX functions mean they will change
some member variables of an object into something specified in
function arguments. But in the *Meter_setValues() case, the new values
are not from the arguments, but from a hard-coded source. The caller
is not supposed to change the values[] to anything it likes, but
rather to "update" the values from the source. Hence, updateValues is
a better name for this family of functions.

### DIFF
--- a/BatteryMeter.c
+++ b/BatteryMeter.c
@@ -32,7 +32,7 @@ int BatteryMeter_attributes[] = {
    BATTERY
 };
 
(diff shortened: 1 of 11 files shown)


hishamhm:
Merged, thank you!

PR #490, opened 2016-05-07
Fix macro Header_forEachColumn

Explorer09:
The (this_) token was not expanded properly, but the bug was not caught
because all uses of this macro specifies (this_)=this .

Also parenthesize macro tokens to prevent further problems.


hishamhm:
Good catch!!

PR #491, opened 2016-05-07
uname meter (Explorer09's implementation of #434 and #448)

Explorer09:
Screenshot

"uname: OS kernel name and release"

Explorer09's implementation of #434 and #448

Caveat: The color (attribute) for uname meter currently shares the one
for hostname meter. This may be changed in the future.


UnameMeter.c
```diff
@@ -0,0 +1,57 @@
...
+}*/
+
+int UnameMeter_attributes[] = {
+   HOSTNAME
+};
+
+static void UnameMeter_updateValues(Meter* this, char* buffer, int size) {
+   (void) this;
+   struct utsname *data = NULL;
+   data = xMalloc(sizeof(struct utsname));
```
GiedriusS: is there a point in using malloc here? why not just:
`
struct utsname data;
uname(&data);
`

Explorer09: @GiedriusS 
Hmm... I need to rethink of this a little bit. My original intention was to have this buffer allocated only once and uname() called only once. But the current Meter object does not allow persistent* storage of non-numeric values yet, so I changed to the allocate-and-call-uname-every-time approach.

So you are suggesting me to allocate this utsname buffer in the stack rather than in the heap, right? That sounds like a good idea, but I doubt whether the code will be portable (you know, `sizeof(struct utsname)` is implementation-defined and it could be quite large, and more likely to cause "overflow" for some systems).

* Persistent within the Meter object's lifetime

PR #494, opened 2016-05-13
Add the "Active users" meter.

nojhan:
The list is truncated to the size of the display.
Color active users according to levels of activity.

Adds ACTIVE_USERS_\* style fields.

This is a fixed and rebased follow-up of pull requests #219, #222 and #264.


Explorer09:
If I'm allowed to comment your patch:
1. I don't like the `Meter.strings[]` idea. Is there any reason that you _have to_ add such variable in the main Meter structure? And you added the same thing in the GraphData structure, why?
2. Your function naming style is unlike the other parts of htop code. Maybe they can be changed right before merging.
3. You did many `malloc` 's and `calloc` 's on function local variables without freeing them. This will result in [memory leaks](https://en.wikipedia.org/wiki/Memory_leak). You better fix them.

nojhan:
Thanks for the review!

3. The e8fd6f0 commit adds the missing free (missed that one, indeed).
2. While I failed to see how the prefixed convention is useful for static functions that are not part of the API, I prefixed all of them in e8fd6f0.
1. I followed the implementation of the load meter, to avoid having different way of doing the same thing across the code base. Unfortunately, the load meter use the `double values[METER_BUFFER_LEN]` in `GraphData` (with a pointer from Meter), which cannot decently be used to store `char*`. So I added an additional `strings` array, using the very same approach than for `values`. I don't think I would risk myself refactoring the general data handling system in this PR, but this could be done later.

Explorer09:
@nojhan Could you please tell us the reason you need this `strings` array, so that I can free up the time understanding your code, and discuss whether there are alternatives to implement the same thing.

Also, may I see a screenshot of what the meter will look like, before I checkout this PR and build ans test for myself?

nojhan:
0.The `strings` array is used for the very same reason that the `values` array is used. AFAI, `values` stores the collected data to be displayed. As `values` is an array of doubles, it cannot be used to store the strings needed by the active users meter. It's the same logic than in the load meter (three data fields with different colors, for instance):
htop_active_users
I suppose that, as it would be better to use UIDs, we can (ab)use `values` and thus do not need a strings array.

0.5. At the time I wrote the patch, I did not found `Process.st_uid` (a dev doc would really be a plus, btw), I will check that.

1-2. So, is it sufficient to iterate over `totalTasks` and check for `percent_cpu`?

3-4. I was (naively) thinking that htop treated differently threads and processes. How do I properly iterate over threads of the same process? Is the CPU% given for the whole system or each core? How do I know on which core a process run?

Explorer09:
1. The `values` array have one more purpose than just "the collected data to be displayed". They are also used for bar or graph plotting! Look at the CPU meters that you often have on the left - the (stacked) bars are drawn using this `values` thing! However, your newly added `strings` array is only meant for temporary buffer storage, which is why I feel it is inappropriate to put it there.
Some meters do not use `values` for a purpose. Take the HostnameMeter for example; it's simply not applicable.
You should not abuse `values` to store anything other than numeric values. Reasons are given above.

2. I have mostly got the idea of what you want to achieve. So if I have time, I might implement a version by myself (as a code exercise :) ). What you want is essentially a (sorted) list of which users currently have processes been running (within the time interval). The list will be ordered by who takes the most of CPU time, to who takes the least time, if not zero. Am I correct?
Just a reminder that this procedure could be expensive. (I cannot think of an algorithm that brings time complexity to lower than O(totalTasks * N * log(N)) where N=number of active users.)

nojhan:
1. I agree that one "should not abuse values to store anything other than numeric values", that's precisely why I added the strings array in the first place. I'm afraid I failed to see the fundamental difference between data used by the CPU meter and data used by the load meter. Obviously, you better understand the htop architecture than me after having read the code of the load meter :-)

2. It seems you correctly understood the description (note that users should not be shown twice in different levels, though, as stated in the PR description). Note that iff speed is your primary objective, I think it's pretty obvious that there is an algorithm in O(k log n). I'm always happy to provide exercises :-)

I would be greatly grateful if you could ping me once you merged your implementation in master, so that I could deploy it on my cluster instead of my current rough implementation.

Explorer09:
@nojhan 
Well, it seems like I've bumped into a technical limit when I try to implement this myself, see #499. Now I can excuse about your `Meter.strings[]` change.

PR #496, opened 2016-05-18
FreeBSD: fix multithreaded CPU% in process list

tcreech:
This fixes an issue on FreeBSD where CPU usage is listed incorrect for multithreaded processes. 

In FreeBSD, the thread ID space is not shared with the process ID space, unlike on Linux. As a result, when htop calls `kvm_getprocs` using `KERN_PROC_ALL`, all individual threads are reported but some threads share the same PID. When htop slurps in everything yielded by `kvm_getprocs`, it erroneously concludes that all threads in a process are duplicates because they share a PID. The result is that only a single thread and its CPU consumption are reported in the process list. (E.g., if LWP 1000 and 1001 in process 4000 are both CPU-bound, htop will report only 100% CPU for PID 4000 rather than 200%.) Note that this does not affect the usage percentages shown by CPU bars.

A quick fix is to use `kvm_getprocs` with `KERN_PROC_PROC` rather than `KERN_PROC_ALL`: this will report only one item per PID, resulting potentially in `ki_pctcpu` > 100. This matches htop's behavior in Linux, except in Linux you can additionally see the individual thread IDs listed.

### DIFF
--- a/freebsd/FreeBSDProcessList.c
+++ b/freebsd/FreeBSDProcessList.c
@@ -424,7 +424,7 @@ void ProcessList_goThroughEntries(ProcessList* this) {
 
    int cpus = this->cpuCount;
    int count = 0;
-   struct kinfo_proc* kprocs = kvm_getprocs(fpl->kd, KERN_PROC_ALL, 0, &count);
+   struct kinfo_proc* kprocs = kvm_getprocs(fpl->kd, KERN_PROC_PROC, 0, &count);
 
    for (int i = 0; i < count; i++) {
       struct kinfo_proc* kproc = &kprocs[i];
@@ -493,10 +493,6 @@ void ProcessList_goThroughEntries(ProcessList* this) {
             if (cpus > 1)
                proc->percent_cpu = proc->percent_cpu / (double) cpus;
          }
-      }
-      if (isIdleProcess == false && proc->percent_cpu >= 99.8) {
-         // don't break formatting
-         proc->percent_cpu = 99.8;
       }
 
       proc->priority = kproc->ki_pri.pri_level - PZERO;


tcreech:
This should fix the issue discovered by @myfreeweb in pull request #427. (See [their comment here](https://github.com/hishamhm/htop/pull/427#issuecomment-189909845)).

This issue pertains to the process list, not the CPU bars.

freebsd/FreeBSDProcessList.c
```diff
@@ -494,10 +494,6 @@ void ProcessList_goThroughEntries(ProcessList* this) {
                proc->percent_cpu = proc->percent_cpu / (double) cpus;
          }
       }
-      if (isIdleProcess == false && proc->percent_cpu >= 99.8) {
-         // don't break formatting
-         proc->percent_cpu = 99.8;
-      }
```
hishamhm: Hi, thank you for the PR!! Could you explain what situation is this test protecting against? Percentages above 100% shouldn't in principle be a problem for formatting.

tcreech: I'm not sure what was supposed to be happening here, but it seems misguided.

I suspect the check for `isIdleProcess` is just an optimization to avoid doing the floating point comparison when unnecessary, and the cap at 99.8 seems unnecessary as you've mentioned. The kernel will actually directly report percentages greater than 100% in the case of multiple threads, and we definitely want to show that.

hishamhm: Oh sorry, I misread the diff and thought you were adding this instead of removing. It makes total sense. I probably need more coffee. :)

etosan:
@tcreech, regarding isIdleProcess, well I think, I added that, when I was to trying to make FreeBSD part. I read their top and tried to replicate it as much as possible, but took liberty with kernel procs. 

Since then few issues with accounting were fixed by good FreeBSD folks.

Here, the thing was two fold:
  - on FreeBSD, kernel is actually reported as process with PID 0.
  - on FreeBSD there are user visible kernel processes (not threads), which have PIDs from normal PID namespace.

This is completely different from Linux behavior, I don't know about other platforms.

Because of this, htop is affected in a few ways:
  - process hierarchy, when in process tree view and kernel threads are set to visible, is rooted in "kernel" (PID0) as "init" is PID1 with PPID0.
  - "intr" kernel process, which is probably handling interrupts, has some weird priority defined.
  - "idle" kernel process was and is still getting insanely large numbers sometimes, for column like TIME+ and sometimes for CPU (as it "runs" most of the "uptime" time if machine is not busy)
  -- on my 8 core boxes, I think, it was getting quite a big combined CPU time which didn't fit into column (not sure if I remember this exactly, but column got broken for sure) 

So I got an idea to treat it as special process and did this hack around it. After all "idle" "eats" all the remaining CPU time and I personally considered +100% as confusing. 

As there were changes, I am not sure, if it goes showing ~800% on 8 core box again, but it would still be great if large numbers option ignored TIME+ stats for "idle" as it makes no sense to color it red (it would always end up red). 

Thus conclusion is, on FreeBSD it's given both "intr" and "idle" are not your "usual" processes and perhaps would need some tiny bit of special treatment.

tcreech:
@etosan, thanks for clearing this up. With this pull, "idle" will indeed show up again with 800% usage on an idle 8-core box.

I would agree that the "idle" process deserves some special treatment. Maybe htop should behave like "top -z", and simply omit the "idle" process in the list? For what it's worth, this is what "atop" on FreeBSD seems to do.

etosan:
I personally am for keeping "idle" around, if given OS does actually have it (BSDs, Windows have actual "idle" entry in proctable). I don't know about OS X. Linux doesn't seem to show it.

myfreeweb:
@etosan OS X doesn't have `idle`, but it has `kernel_task` that shows time spent in kernel.

hishamhm:
I'd rather not need special cases for the idle process in the cross-platform parts, so I'd prefer to hide idle in systems that having, and keep the invariant that CPU% total is the sum of the CPU% of all processes.

Having `kernel_task` is fine because it is like the various kernel threads that show up in Linux.

etosan:
Well this is in platform parts anyway, and on FreeBSD 'idle' and 'intr' are guaranteed to be there,  moreover they are both "kernel processes" as well.

hishamhm:
@etosan this would cause the CPU meter graphs in the cross-platform parts to add up to 100% all the time

etosan:
Roger, whatever floats the boat.

PR #500, opened 2016-05-27
Remove redundant is-null checks on free(Meter.drawData)

Explorer09:
Unused Meter.drawData pointers shall be NULL by the virtue of using calloc() in initializing the Meter objects. And `free(NULL)` is harmless. So these conditionals make no sense.


hishamhm:
Thanks!

PR #504, opened 2016-06-02
Fixes upon xStrdup(NULL) problem

Explorer09:
The "Stricter strdup." 4674b4a commit is wrong. Checking whether the pointer argument is NULL helps nothing but increased code size. It would crash anyway on NULL no matter it has checking code or not.
This pull request reverts the commit, but also adds proper code to check xStrdup(NULL) bugs.
Namely, assertions (in macros) and `__attribute__((nonnull))`.


hishamhm:
thank you!!

PR #510, opened 2016-06-16
Implement option to collapse all entries in tree view

### LINKED ISSUE #24: Implement option to collapse all entries in tree view

chaitan94:
Added the ability to toggle (expand / collapse) all the trees in the tree view (by pressing the key 'E').

The preference is also made persistent using the settings, and can also be modified from Display Options menu.

I think this solves the issue #24

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -224,6 +224,15 @@ static Htop_Reaction actionToggleTreeView(State* st) {
    return HTOP_REFRESH | HTOP_SAVE_SETTINGS | HTOP_KEEP_FOLLOWING | HTOP_REDRAW_BAR | HTOP_UPDATE_PANELHDR;
 }
 
+static Htop_Reaction actionExpandOrCollapseAllTrees(State* st) {
+   st->settings->allTreesCollapsed = !st->settings->allTreesCollapsed;
+   if (st->settings->allTreesCollapsed)
+      ProcessList_collapseAllTrees(st->pl);
+   else
+      ProcessList_expandTree(st->pl);
+   return HTOP_REFRESH | HTOP_SAVE_SETTINGS;
+}
+
 static Htop_Reaction actionIncFilter(State* st) {
    IncSet* inc = ((MainPanel*)st->panel)->inc;
    IncSet_activate(inc, INC_FILTER, st->panel);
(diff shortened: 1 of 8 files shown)


chaitan94:
Correct me if I'm wrong, but doesn't the label 'feature request' apply to issues which are for requesting the feature? This pull request actually implements the feature. (The title sounds a bit misleading, I know)

hishamhm:
A feature request including code to implement is still a request for a feature to be added. :)

It is definitely not a "fix": a program is not broken just because it doesn't include a feature.

I try to be careful when adding features and taking up new keybindings; there is a hard-to-explain balance that I try to maintain between being featureful and keeping it small.

Since you took the time to implement it, I'll take the time to try to explain my reasoning for not having added it. Collapsing the whole tree down to `init` doesn't sound useful, so I guess one would only use this to uncollapse part of it straight-away. So this means the user probably wanted to collapse only part of the tree, which seems indicative of a UI need that can probably be addressed more efficiently some other way.

hishamhm:
Having said that, sorry for not using the code. It's always a bummer to have a PR rejected, but thank you for taking the time and diving into the code! (I hope it was easy to follow!)

v1nc:
Why is this rejected?
You could change the keybinding or only make the tree view collapsed at default.
The tree view is very useless by now. Some applications start hundreds of processes. If I do not find the main application in the normal view by luck, I need the tree view to kill only the parent application, so I do not have to kill hundreds of processes by hand. If the tree view would be collapsed by default, or there would be a keybinding for it, I could do my task in some seconds. Without it, I actually have to collapse all processes by hand. This is really annoying!
I do not understand why you reject these pulls and just ignore the problem. People created issues here that pointed out the problem and even after someone else pushes a fix, you ignore the problem.

To say it clear: This PR **is** definitely **a fix**! This is not a missing feature, the tree view is broken and nearly useless by now.

SriramKid:
I would really like this change to be added. I usually have htop open in my 1st workspace, and having to collapse all the trees to get an overall view of what processes are open is quite a pain. This would make the presentation very pleasant on opening, and when needed an uncollapse all can be used for all other purposes.

PR #512, opened 2016-06-17
Man page fix: Write '+' for key combinations (Ctrl+ or Alt+)

Explorer09:
Be consistent. Most other GUI applications use '+' for key combinations joiner, only a few use '-'.


hishamhm:
This is a matter of taste. There's no hard rule on this; I've been using "-".

Issue #513, opened 2016-06-18
Basename problem with some Google Chrome processes


Explorer09:
Some Google chrome processes may not show or highlight the correct "basename" in the Command field. Let me show an example with some screenshots:
![htop screen, without 'Show program path'](https://cloud.githubusercontent.com/assets/2487263/16168039/82656a10-3530-11e6-993e-580b352e49fa.png)
![htop screen, with 'Show program path'](https://cloud.githubusercontent.com/assets/2487263/16168045/8c2c6a3a-3530-11e6-8f06-ba34a5c8ce3a.png)

But in `top`, the names are shown correctly ("chrome"):
![top screen](https://cloud.githubusercontent.com/assets/2487263/16168059/bf58b86e-3530-11e6-8096-0f664e9d4380.png)

According to my research, this is not completely htop's fault. But Chrome sometime changes its `cmdline` entry so that its first part (the string before the first `\0` delimiter) concatenates file names and some of its options.

For reference, here's the `cmdline` entry for process 5443 as shown in the example:

![proc/5443/cmdline output](https://cloud.githubusercontent.com/assets/2487263/16168124/a60bd138-3531-11e6-9805-71e2ea441702.png)

    $ tr '\0' '\n' </proc/5443/cmdline | cat -n
[...]

PR #519, opened 2016-06-22
wrap command line in a separate screen

ryenus:
For certain process, the command line could be very long, especially for Java programs that having many command line options.

It would be quite helpful to be able to see the full command line, wrapped into multiple lines in a separate window, so that one can inspect all the command line options conveniently.

Please refer to the screenshot in the comment below to see it in action.


ryenus:
A screenshot for a running process of ActiveMQ

image

This should make it much easier to see every piece of the command line of any process.

Explorer09:
According to your screenshot, may I ask a few questions before I checkout the code and test it:
Do you wrap lines only when multiple arguments do not fit into one line for display, or do you wrap lines on every argument?
Actually my original suggestion is for the latter behavior. Because it'll not only look more elegant but it'll be easier to distinguish "embedded" spaces and dashes from real option specifiers (like finding this bug: #513).

Also, it will be a bonus if you can display special characters (especially `\n` ) is escaped form. This would make the strings safer to print out and reuse: Example:

    # When invoked in shell like this:
    echo -n -e '\033'[33myellow \'text\''\n'

When parsed in htop, it shall be displayed in this safe form (note the special characters all being escaped and text is single-quoted):

    echo
    -n
    -e
    '\x1B[33myellow '\'text\''\n'

ryenus:
My motivation is to deal with long command lines which are too long to inspect easily, therefore I want to make the selected long command line fully displayed in one screen, so a line is wrapped only when it doesn't have enough room for the next argument.

Since all we get is `this->process->comm`, which already has some quotes stripped, and concatenated into one string, I agree with you that we probably can do something regarding escaping, but it might not be possible to completely restore the original input, at least in its exact form.

And, regarding the escape sequence, Could you please elaborate the security assumption and/or concern when you say this? 
> make the strings safer to print

Thank you

ryenus:
@Explorer09 I agree with you that it would be really cool to be able to handle those special characters, especially escape sequences, and present them in a nice way. I'd also like to find a solution for it, though I'm not sure about the technical feasibility. 

For now the goal of making the full command line easily viewable, especially targeting processes with rather long command line and many command line options, typically java processes, IMHO, is well served. And it's also searchable, and filterable.

Honestly speaking, regarding special characters handling, maybe you or someone interested can come up with a solution, or maybe later?

Explorer09:
@ryenus 
Umm. I'm sorry. Helping you write an algorithm to handle special characters that isn't my priority now.
I have no urge to use this feature of htop you provide. And, AFAIK, the problem of not handling special characters isn't that severe as you can expect --- the most severe we can have is the corruption of visuals on screen [like this XKCD comic](https://www.explainxkcd.com/wiki/index.php/1137:_RTL). So when I perceive that can be done, it's not a priority for me.

ryenus:
Thanks, and no worries, actually I wanted to say it isn't a priority either, regarding:
> And, AFAIK, the problem of not handling special characters isn't that severe as you can expect

That's why I said maybe someone interested can take a chance on it when the time is right. Thanks

PR #524, opened 2016-07-10
Allow to override ncurses*-config path

ricardo0y:
This will be used when cross-compiling with ncurses*-config generated for the
target, using constructs like
htop_ncurses_config_script=/path/to/ncurses5-config

Signed-off-by: Ricardo Martincoski <ricardo.martincoski@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -147,7 +147,12 @@ fi
 # HTOP_CHECK_SCRIPT(LIBNAME, FUNCTION, DEFINE, CONFIG_SCRIPT, ELSE_PART)
 m4_define([HTOP_CHECK_SCRIPT],
 [
-   htop_config_script=$([$4] --libs 2> /dev/null)
+   if test ! -z "m4_toupper($HTOP_[$1]_CONFIG_SCRIPT)"; then
+      # to be used to set the path to ncurses*-config when cross-compiling
+      htop_config_script=$(m4_toupper($HTOP_[$1]_CONFIG_SCRIPT) --libs 2> /dev/null)
+   else
+      htop_config_script=$([$4] --libs 2> /dev/null)
+   fi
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -147,7 +147,11 @@ fi
 # HTOP_CHECK_SCRIPT(LIBNAME, FUNCTION, DEFINE, CONFIG_SCRIPT, ELSE_PART)
 m4_define([HTOP_CHECK_SCRIPT],
 [
-   htop_config_script=$([$4] --libs 2> /dev/null)
+   if test ! -z "$htop_[$1]_config_script"; then
```
hishamhm: please make the name of this variable all-uppercase (`as in HTOP_[$1]_CONFIG_SCRIPT`). the htop_* versions were meant to be "local" variables for the macro. Adding a short comment here explaining this was motivated by cross-compiling scenarios would also be welcome, for future maintenance.

configure.ac
```diff
@@ -147,7 +147,12 @@ fi
 # HTOP_CHECK_SCRIPT(LIBNAME, FUNCTION, DEFINE, CONFIG_SCRIPT, ELSE_PART)
 m4_define([HTOP_CHECK_SCRIPT],
 [
-   htop_config_script=$([$4] --libs 2> /dev/null)
+   if test ! -z "m4_toupper($HTOP_[$1]_CONFIG_SCRIPT)"; then
```
ricardo0y: I used
m4_toupper($HTOP_[$1]_CONFIG_SCRIPT)
because I think it is more readable, but I can change to
$HTOP_[]m4_toupper($1)_CONFIG_SCRIPT
if you want

hishamhm: looks fine!

PR #528, opened 2016-07-24
Optimize GraphMeterMode_dots (now GraphMeterMode_styles)

Explorer09:
An attempt to optimize GraphMeterMode_dots from a array of strings into
a single string, removing the need of a bunch of pointers (to string
literals of each Braille code point). Now it uses padding instead.

The file-scope GraphMeterMode_pixPerRow and GraphMeterMode_dots
variables are now converted into local ones in GraphMeterMode_draw(),
for smaller code.

Technical note: The script MakeHeader.py does not (yet) allow me to
write a file-scope variable definition that span multiple lines without
enclosing definitions in braces {}, so GraphMeterMode_styles declared
as an array of strings is my way to workaround that script limit.
(The extensibility of GraphMeterMode_styles to coding more style in was
not part of original intention, but we have it anyway.)

```
$ gcc --version | head -n 1
gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4
$ uname -m
x86_64
$ size htop.old htop.new
   text   data   bss     dec    hex filename
 137929  15112  3776  156817  26491 htop.old
 137993  14792  3776  156561  26391 htop.new
```

Signed-off-by: Kang-Che Sung

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -333,28 +333,58 @@ static void BarMeterMode_draw(Meter* this, int x, int y, int w) {
 
 /* ---------- GraphMeterMode ---------- */
 
-#ifdef HAVE_LIBNCURSESW
+/*{
+typedef enum {
+   GRAPHSTYLE_ASCII = 0, // Fallback default
+   GRAPHSTYLE_UTF8 = 1,  // UTF-8 default (braille)
+   LAST_GRAPHSTYLE       // Dummy & unused.
+} GraphStyleId;
+}*/
 
-#define PIXPERROW_UTF8 4
-static const char* GraphMeterMode_dotsUtf8[] = {
-   /*00*/" ", /*01*/"⢀", /*02*/"⢠", /*03*/"⢰", /*04*/ "⢸",
-   /*10*/"⡀", /*11*/"⣀", /*12*/"⣠", /*13*/"⣰", /*14*/ "⣸",
-   /*20*/"⡄", /*21*/"⣄", /*22*/"⣤", /*23*/"⣴", /*24*/ "⣼",
-   /*30*/"⡆", /*31*/"⣆", /*32*/"⣦", /*33*/"⣶", /*34*/ "⣾",
-   /*40*/"⡇", /*41*/"⣇", /*42*/"⣧", /*43*/"⣷", /*44*/ "⣿"
+static const char *const GraphMeterMode_styles[] = {
+   /* 1st byte: num of quantization levels per row per character. (pixPerRow)
+    * 2nd byte: num of bytes per multibyte character. Every character to be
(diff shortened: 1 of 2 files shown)


Explorer09:
In case that @eworm-de is interested, here's a how a "block elements" style may be implemented after this pull request (note, however, this patch does not include a way to change graph meter style in runtime, so you will gonna overwrite the default one if you want to use this style):

    [GRAPHSTYLE_BLOCK_ELEM] = ( /* pixPerRow */ "\2" /* size */ "\4"
    /*00 [ ]       01 [▗]        02 [▐]*/
    " \0\0\0"     "\xe2\x96\x97\0\xe2\x96\x90\0"
    /*10 [▖]       01 [▄]        02 [▟]*/
    "\xe2\x96\x96\0\xe2\x96\x84\0\xe2\x96\x9f\0"
    /*20 [▌]       21 [▙]        02 [█]*/
    "\xe2\x96\x8c\0\xe2\x96\x99\0\xe2\x96\x88"),

eworm-de:
Thanks for the heads up! ;)
Actually I used the block element version for a very short time only when my bitmap font did not support braille characters. As you did the work already I will update my branch if your change is merged in case anybody is still interested.

hishamhm:
Thanks for the patch! 
Could the elements of GraphMeterMode_styles be made into a struct with fields `pixPerRow`, `size` and `data`? It should produce the same memory layout, and would make the code a bit more readable.

Explorer09:
@hishamhm 
No, the memory layout will not be the same. Converting them into a struct adds a pointer to the string literal which is a waste of at least 8 bytes (assume a 64-bit system). I have considered that already, but decided that concatenating the numeral values with the string literal will be (a bit) smaller when stored in memory.

    +---(4 bytes)---+
    | pixPerRow     |
    +---------------+
    | size          |
    +---------------+        +------------------
    | data          |------> | string literal
    | (pointer)     |        +------------------
    +---------------+

hishamhm:
Yeah, packing a struct with a flexible array member inside another array sounds like something that wouldn't work. The easy way out would be to get rid of the outer array. Just for fun, I fetched your graph-style-test-1, and this is what I came up with:

https://gist.github.com/hishamhm/c1b5badaf2cd03a1cdcce008d93c8606

-Wpedantic still complains, though, because apparently this isn't standard C: http://stackoverflow.com/questions/27852062/why-static-initialization-of-flexible-array-member-work

Anyway, while this was a fun exercise, I think the gains of micro-optimizations such as this are not worth the readability cost (thinking long term maintenance of the code here). I prefer to keep things simple (for my personal definition of simple :) ), and this strategy has paid off so far (and the fact that others find the codebase approachable seems like a good sign!)

Issue #480, opened 2016-04-21
Pressing "Page Down" hides all entries except the last


Sworddragon:
I'm using htop 2.0.1 and after upgrading from version 1.0.3 I'm noticing that pressing "Page Down" in any menu that supports it (like the process list) the selection moves succesfully to the last entry but instead of showing the remaining entries above it until there is no space they are directly moved "offscreen" and thus hidden.

hishamhm:
Fixed, thanks for the push! This always annoyed me slightly but not enough to do anything about it.

PR #534, opened 2016-08-07
(Fix #532) Fix scrolling behavior change

### LINKED ISSUE #532: "Page Down" and "Page Up" work a bit different
I'm using htop 2.0.2 and for example if I'm within a small view of the process list, selecting the first process of it and then pressing "Page Down" the last process of the current view gets selected and if I'm then pressing "Page Up" the first process of the current view gets selected but additionally the view gets moved 1 up. I think either pressing "Page Down" should cause the same shift of the view or on pressing "Page Up" the shift should not appear there anymore to make the behavior of these keys consistent.

Explorer09:
Commit "Make PgDown behavior more usual." 759caf0f8fa593430adea676fc64612b5197dca8
silently changes the PageDown scrolling behavior that, instead of
scrolling one window down until the end of the window touches the end
of the list, the window simply repositions itself in a way that the
selected item always become the last item in the new window.

The commit reverts the behavior, and also fixes sanity conditionals
so that the scrollV variable will _never_ become negative or out-of-
bound.

Fixes issue #532. Also keep the problem #480 addressed.

### DIFF
--- a/Panel.c
+++ b/Panel.c
@@ -407,12 +407,13 @@ bool Panel_onKey(Panel* this, int key) {
       break;
    case KEY_PPAGE:
       this->selected -= (this->h - 1);
-      this->scrollV -= (this->h - 1);
+      this->scrollV = MAX(0, this->scrollV - this->h + 1);
       this->needsRedraw = true;
       break;
    case KEY_NPAGE:
       this->selected += (this->h - 1);
-      this->scrollV = MIN(MAX(0, Vector_size(this->items) - this->h), this->selected - this->h);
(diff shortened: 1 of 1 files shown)


hishamhm:
Thanks! While we're doing this, could you change it to use CLAMP?

Explorer09:
@hishamhm No, I'm afraid that CLAMP macro won't fit here. Look at the logic carefully.
Both `(Vector_size(this->items) - this->h)` and `(this->scrollV + this->h - 1)` could possibly be negative. This isn't the case that CLAMP should work with. (`CLAMP(val,low,high)` must ensure that low<=high without any more conditional, or otherwise it's weird behavior.)

PR #535, opened 2016-08-10
Reject (alpha|sparc|mips|parisc)-linux* build targets.

Explorer09:
Signal tables for these architectures differ from other common ones
(x86, arm and others) in Linux.
Although it will be better to implement signal tables for these, we
reject these platforms for now until someone can work on implementing
that.

Note that configure options like
'--host=alpha-linux-gnu --target=i686-linux-gnu'
should continue to work, so that people who are ambitious can still
build it anyway, circumventing our target check.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -48,6 +48,15 @@ darwin*)
    ;;
 esac
 
+dnl Signal tables for these architectures differ from other common ones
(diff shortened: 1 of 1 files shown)


hishamhm:
Event with a wrong symbol table, it's useful to have the program available in these architecture. Having something is better than nothing.

PR #538, opened 2016-08-12
2 minor Vector.c tweaks

Explorer09:
2 minor Vector.c tweaks that might be considered merging and should result in no changes in the binary (for release builds).


hishamhm:
I think swap() is short/simple enough to be a macro, and I prefer that Vector_get remains a function in debug builds, so it appears explicitly in backtraces. For this reasons, I prefer to keep the code as originally designed.

PR #546, opened 2016-08-26
Fix typos

Gelma:


Explorer09:
The one in linux/LinuxProcessList.c isn't a typo. It's intentional. Please revert, or do a favor to change it to something like:

    if (strncmp(line+1, "cancelled_write_bytes: "+1, 22) == 0) {
    // note the +1 after the string literal

GiedriusS:
Indeed it's not a typo. I think what would be nicer there instead is to use sscanf() to parse the lines.  Currently how LinuxProcessList_readIoFile() looks is terrible.

hishamhm:
@GiedriusS it's done that way because it's noticeably faster. Reading the process list is performance-critical.

hishamhm:
@Gelma please keep the one in LinuxProcessList.c as is.

Gelma:
Hi Hishamhm,
   thanks a lot for your review.
   It's a mess in these days. So, please, feel free to merge my "fixes" as you prefer. I don't care if you don't use my commit, or if the commit author is you. No problem.

Thanks a lot again,
Andrea

hishamhm:
@Gelma all right! Merged, thank you!

PR #547, opened 2016-08-26
Ported my network traffic meter I made years ago for htop 0.9

cobrafast:
This is a rather crude implementation, based on what I had done in an older project of mine called "ahtop".

Things that are missing (or not 100% there):
- Meter does not adjust to refresh duration / pause length (whatever you call it), meaning anything but 1 second refreshes will lead to wrong readings (I don't know where to get the refresh duration from)
- Meter is no longer switchable between 100 Mbps / 1 Gbps and is fixed to 1 Gbps (I had this in the version for htop 0.9, I should be able to port this as well if necessary)
- Only implemented for linux platform (I don't know if it breaks other platforms the way it's right now and I don't know how to implement it for other platforms)

I was hoping that someone with more experience in C and this particual project would appreciate this effort and be able to get it up to speed relatively easily.

If there's anything I can do to improve it let me know.


Explorer09:
I didn't test it yet, but I can give some programming hints:
1. You are free to adapt the Meter_humanUnit function to your own need; that would allow more units than Mb/s and Gb/s switch.
2. I have my own graph meter dynamic scaling feature under development (see pull request #487), and with it a more dynamic bar drawing can be achieved. You may consider it.

amosbird:
Hi, is there any plan to resurrect this PR?

PR #549, opened 2016-08-29
Linux signal tables for Alpha, SPARC, MIPS & PARISC

Explorer09:


hishamhm:
Hi, did you stumble upon any of these machines?

Explorer09:
@hishamhm  No, I don't have any of them. What I cared is that the [Linux signal man page](http://man7.org/linux/man-pages/man7/signal.7.html) already stated that some architectures have different signal numbers, and I would consider that a bug if we don't support it or at least detect it at configure time.

Note that many `kill` utilities (from bash, procps, or busybox) don't have this signal table problem because they grab the signal numbers from the system headers (at compile time), which glibc has already taken care of (see `sysdeps/unix/sysv/linux/<arch>/bits/signum.h` in its source code). Htop uses hard-coded tables instead of fetching numbers from libc, so we should include tables for all architectures.

PR #550, opened 2016-08-30
Mark signal tables 'const'

Explorer09:
Specifically, Platform_signals[] and Platform_numberOfSignals. Both are
not supposed to be mutable. Marking them 'const' puts them into rodata
sections in binary. And for Platform_numberOfSignals, this aids
optimization (aids only Link Time Optimization for now). :)


hishamhm:
Merged, thanks! :)

PR #551, opened 2016-08-30
Real-time signals support (kill command)

Explorer09:
SignalsPanel_new now fetches SIGRTMIN and SIGRTMAX and generates real-
time signals entries at runtime.

### DIFF
--- a/SignalsPanel.c
+++ b/SignalsPanel.c
@@ -31,13 +31,36 @@ Panel* SignalsPanel_new() {
    Panel* this = Panel_new(1, 1, 1, 1, true, Class(ListItem), FunctionBar_newEnterEsc("Send   ", "Cancel "));
    const int defaultSignal = SIGTERM;
    int defaultPosition = 15;
-   for(unsigned int i = 0; i < Platform_numberOfSignals; i++) {
+   unsigned int i;
+   for (i = 0; i < Platform_numberOfSignals; i++) {
       Panel_set(this, i, (Object*) ListItem_new(Platform_signals[i].name, Platform_signals[i].number));
       // signal 15 is not always the 15th signal in the table
       if (Platform_signals[i].number == defaultSignal) {
          defaultPosition = i;
       }
    }
+   #if (defined(SIGRTMIN) && defined(SIGRTMAX))
+   // Real-time signals.
+   // SIGRTMIN and SIGRTMAX expand to libc internal functions and we have to
+   // grab their numbers at runtime.
+   static char buf[15]; // 15 == sizeof("xx SIGRTMIN+nn")
+   int rtmax;
+   for (int sig = SIGRTMIN; sig <= (rtmax = SIGRTMAX); i++, sig++) {
+      // Every signal between SIGRTMIN and SIGRTMAX are denoted in "SIGRTMIN+n"
+      // notation. This matches glibc's strsignal(3) behavior.
+      // We deviate from behaviors of Bash, ksh and Solaris intentionally.
+      if (sig == rtmax) {
+         snprintf(buf, 15, "%2d SIGRTMAX", sig);
+      } else {
+         int n = sig - SIGRTMIN;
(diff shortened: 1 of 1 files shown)


hishamhm:
is it common to use `SIGRTMAX-n`? It seems to me that `SIGRTMIN+n` sounds more useful, since it's a zero-based index of all RT signals. I'm considering merging this variant:

```c
   #if (defined(SIGRTMIN) && defined(SIGRTMAX))
   // SIGRTMIN and SIGRTMAX expand to libc internal functions and we have to
   // grab their numbers at runtime.
   static char buf[15];
   for (int sig = SIGRTMIN; sig <= SIGRTMAX; i++, sig++) {
      int n = sig - SIGRTMIN;
      if (sig == SIGRTMAX) {
         snprintf(buf, 15, "%2d SIGRTMAX", sig);
      } else {
         snprintf(buf, 15, "%2d SIGRTMIN%-+3d", sig, n);
         // Bash's builtin "kill" does not accept "SIGRTMIN+0"; must specify
         // "SIGRTMIN" without "+0" notation.
         if (n == 0) {
            buf[11] = '\0';
         }
      }
      Panel_set(this, i, (Object*) ListItem_new(buf, sig));
   }
   #endif
```

(Note also "SIG" in the signal label)

Explorer09:
For the `SIGRTMAX-n` part, I only tried to mimic the output of what bash's or dash's `kill -l` gives me. And guess what, `kill -l RTMIN+16` gives me `bash: kill: RTMIN+16: invalid signal specification` (assuming SIGRTMIN==34 and SIGRTMAX==64), so I seem to me that numbering all RT signals from RTMIN up is a bad idea.

By the way, [this thread in bash mailing list](https://lists.gnu.org/archive/html/bug-bash/2013-05/msg00039.html) might be worth reading.

Explorer09:
Patch revised. Added SIG prefix. Simplify format string a little bit.
I'm still not convinced about removing SIGRTMAX-n things.

hishamhm:
That mailing list thread that was linked mentions that SIGRTMIN+n is a conceputally more stable way of referring to a signal number.

Explorer09:
Yes, "conceptually" only. One of the issue is that there is no guarantee that, for example, SIGRTMIN+16 will be always available (i.e. less than or equal SIGRTMAX) in all environments. And for the info, POSIX.1-2008 only mandates that a minimum of 8 RT signals (_POSIX_RTSIG_MAX) should be supported if a system supports RT signals (references: [[signal.h]](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html) [[limits.h]](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html)).

**My primary concern is that, for now, no shells ever implement signal tables (the `kill -l`) with everything denoted in SIGRTMIN+n style.** And so htop shouldn't the first one that breaks the (de facto) convention of that.

hishamhm:
Many of those "de facto conventions" are just developers shying away of making decisions and justifying a design by saying "I'm doing this way because someone else did it like this too". Htop was created out of desire of doing things differently than other software; it wouldn't do half of things it does if "following de facto conventions" was in its design criteria. I do it when I think it makes sense. (And I've been around long enough to see these "de facto conventions" change over time... some of them, amusingly, influenced by htop.)

In the case of Bash it is justifiable that it copies other shells' behaviors because of because of different softwares are installed in different OSes as /bin/sh, so compatibility with legacy platforms is a matter of interoperability. Htop, however, is not a shell and does not aim to be "compatible" with any other software (i.e. it does not purport itself to be a "top", like many different programs that share this same name).

So no, it's not a decisive factor for the design of htop whether Bash or /usr/xpg4/bin/sh on SunOS5 from 1992 do something a particular way.

Explorer09:
Revised (again). All signals between SIGRTMIN and SIGRTMAX are written in "SIGRTMIN+n"
notation now.

hishamhm:
Thanks!

Issue #314, opened 2015-11-29
Rationale behind "RT" value in prio column


abusque:
Hi,

I've been working with task priorities on Linux quite a bit over the last few days, and I used htop to monitor my changes most of the time. During that time, I noticed that htop displays "RT" in the priority column only when the task priority reaches -100. See [`Process.c:427`](https://github.com/hishamhm/htop/blob/master/Process.c#L427). I was wondering what's the rationale behind this, given that, from the kernel's point of view at least, all negative task priorities correspond to real-time priorities. This can be seen at [`include/linux/sched/rt.h:6`](https://github.com/torvalds/linux/blob/master/include/linux/sched/rt.h#L6) in the kernel's source tree. Keep in mind that the priority seen by htop (or through `/proc`) is offset by `MAX_RT_PRIO`.

So, I can understand not wanting to print "RT" for all negative priority values, as this would hide too much information, but then, why put it only for a -100 task priority, instead of not at all?

hishamhm:
> So, I can understand not wanting to print "RT" for all negative priority values, as this would hide too much information, but then, why put it only for a -100 task priority, instead of not at all?

This code is really old (I traced the "RT" line back to the initial import of this git repository, from back in 2006!) and I don't remember exactly what the rationale was, to be honest... I couldn't find anything in my email archives, either. I _think_ I just copied the behavior of the version of top that was current back in the day. It certainly might be worth revisiting.

abusque:
That was my impression from a quick `git blame` as well. A quick look at procps-ng's top (which is the version currently shipped by Arch Linux, and seems to have superseded the original procps in most major distros) show similar behaviour. See [`top/top.c:5484`](https://gitlab.com/procps-ng/procps/blob/master/top/top.c#L5484), or [`top.c:3123`](http://procps.cvs.sourceforge.net/viewvc/procps/procps/top.c?view=markup) for the original procps.

The difference is that they display "RT" for all values below -99, not just -100, which I think correctly maps to task priorities from (much) older kernels.

I think the options would then be:

* Retain the current behaviour for "compatibility" with other top implementations, although I'm not sure how much of a factor this is for htop.
* Not displaying "RT" for any priority value, though this increases the user's cognitive load by requiring them to know that negative priorities are for real-time tasks.
* Displaying "RT" for all negative priority valuee. I think this option hides too much useful information.
[...]

PR #555, opened 2016-09-06
Fix column misalignment for priority -101 threads

ivan:
BFS-patched kernels can have kernel threads with priority -101.
This change makes priority -101 display as "RT", just like priority -100.

Related: https://github.com/hishamhm/htop/issues/314

screenshot_2016-09-06_09-18-54


ivan:
Also, I am not 100% sure this is related to BFS, it could just be PREEMPT.

hishamhm:
Merged, thank you!

PR #558, opened 2016-09-17
Limit signal list display to first 99 signals.

Explorer09:
(Apology if I'm impolite in making yet another pull request for this. But I simply want to make htop code better.)

This reverts "Protect against strange values of SIGRTMIN and SIGRTMAX,
and change alignment of numbers." (5df425867cd992ccadbe8b5cf713404fd53db7df)
and changes it to a (I believed) better bound protection. Now the list
only displays first 99 signals that are available.

In case of a system providing so called "strange" or too large SIGRTMAX
value, let it be. That may not be an error but an OS feature.

Also note the "%+d" to "%-+3d" format change was technically redundant.

### DIFF
--- a/SignalsPanel.c
+++ b/SignalsPanel.c
@@ -40,16 +40,26 @@ Panel* SignalsPanel_new() {
       }
    }
    #if (defined(SIGRTMIN) && defined(SIGRTMAX))
-   if (SIGRTMAX - SIGRTMIN <= 100) {
-      static char buf[15];
-      for (int sig = SIGRTMIN; sig <= SIGRTMAX; i++, sig++) {
+   // Real-time signals. SIGRTMIN and SIGRTMAX may expand to libc internal
+   // functions and so grab their numbers at runtime.
+   static char buf[15];
+   for (int sig = SIGRTMIN; sig <= 99; i++, sig++) {
+      // Every signal between SIGRTMIN and SIGRTMAX are denoted in "SIGRTMIN+n"
+      // notation. This matches glibc's strsignal(3) behavior.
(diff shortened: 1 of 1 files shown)


hishamhm:
Thank you, but there is no need for this; in modern OSes, signal tables are bitfields, so they are either 32 or 64 bit in size. I only added the check because the glibc source code mentioned that some OSes return bogus values for SIGRTMIN and MAX. I put that in place to avoid a crash in the odd chance someone running something like AIX attempts a port.

Explorer09:
@hishamhm Thank you, but you didn't convince me about your code change. I now see that some OSes don't give us a useful SIGRTMIN and SIGRTMAX value (from what I found out on the web, "AIX 4.3 defines SIGRTMIN and SIGRTMAX as 888 and 999 respectively"), but isn't that better to check this thing at the preprocessor time, so that no redundant code would be generated for compliant systems such as Linux?

    // Guard against bogus SIGRTMIN and MAX values
    // (HP-UX 10.20 defines both SIGRTMIN and SIGRTMAX as -1;
    // AIX 4.3 and 5.1 define SIGRTMIN and SIGRTMAX as 888 and 999 respectively)
    // In case that SIGRTMAX expands to __libc_current_sigrtmax(), preprocessor
    // would evaluate this to 0, which helps the trick.
    #if defined(SIGRTMAX) && SIGRTMAX >= 0 && SIGRTMAX < 100

And also add an assertion in case of runtime functions return a bogus value (definitely a bug that the OS/libc team should be responsible for):

    assert(SIGRTMAX > 0 && SIGRTMAX < 100);

Won't this be better?

PR #579, opened 2016-11-22
Improve /proc/tty/drivers scanner

Explorer09:
Smaller code. And I'm trying to utilize the printing format of
/proc/tty/drivers to make parsing a little bit faster (in theory only,
didn't actually measure the performance).

    $ size htop htop.old
       text   data    bss    dec    hex	filename
     140543  14560   3808 158911  26cbf	htop
     140671  14560   3808 159039  26d3f	htop.old

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -150,34 +150,31 @@ static void LinuxProcessList_initTtyDrivers(LinuxProcessList* this) {
    ttyDrivers = malloc(sizeof(TtyDriver) * allocd);
    char* at = buf;
    while (*at != '\0') {
-      at = strchr(at, ' ');    // skip first token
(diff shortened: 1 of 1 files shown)


hishamhm:
Thank you, but I'd rather keep my own code which makes less assumptions about the format.

Explorer09:
@hishamhm 
Technically, there is only one additional assumption: that the first field is padded with spaces and at least 20-chars wide. The other assumptions are exactly the same as what your code have assumed. Really.

And note that the atoi() and strtol() family of functions skip leading whitespaces so you don't need to skip them yourself.

If you wish, I can tweak the code a little bit so that the assumptions exactly match what your code has. Will it be possible to pull then?

hishamhm:
No thanks, lately I'm very short in time for htop. I'd rather spend the litle time I have on bugs.

PR #584, opened 2016-12-29
Replace all uses of sprintf with snprintf

EliteTK:
In all the cases where sprintf was being used within htop, snprintf
could have been used. This patch replaces all uses of sprintf with
snprintf which makes sure that if a buffer is too small to hold the
resulting string, the string is simply cut short instead of causing
a buffer overflow which leads to undefined behaviour.

`sizeof(variable)` was used in these cases, as opposed to `sizeof
variable` which is my personal preference because `sizeof(variable)`
was already used in one way or another in other parts of the code.

### DIFF
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -48,7 +48,7 @@ static void CPUMeter_init(Meter* this) {
    int cpu = this->param;
    if (this->pl->cpuCount > 1) {
       char caption[10];
-      sprintf(caption, "%-3d", Settings_cpuId(this->pl->settings, cpu - 1));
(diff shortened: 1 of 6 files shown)


hishamhm:
Hi! As you probably noticed, all uses of `sprintf` were only in statically-allocated buffers, with format strings that expand to bounded sizes (e.g. no uses of `%s`). So I believe there are no chances of buffer overruns with current uses of `sprintf` in the codebase.

Still, it shouldn't hurt to switch them to `snprintf`, and it does make the code a bit less fragile.

hishamhm:
Merged, thank you!

PR #592, opened 2017-01-31
CPU clock rate can be displayed in meter instead of CPU load percentage

aleksei-udalov:
Modern CPU can change frequency in wide range. I think it may be useful just to see the freq in htop.

This feature is disabled by default and can be enabled in setup->display options. This commit affected only text in meter. Bar keeps show CPU load. Tested on Linux only.

5v1br_croper_ru

### DIFF
--- a/CPUMeter.c
+++ b/CPUMeter.c
@@ -63,7 +63,11 @@ static void CPUMeter_updateValues(Meter* this, char* buffer, int size) {
    }
    memset(this->values, 0, sizeof(double) * CPU_METER_ITEMCOUNT);
    double percent = Platform_setCPUValues(this, cpu);
-   snprintf(buffer, size, "%5.1f%%", percent);
+   if (this->pl->settings->showClockRate) {
+      snprintf(buffer, size, "%5.1lfMHz", this->clockRate);
+   } else {
+      snprintf(buffer, size, "%5.1f%%", percent);
+   }
 }
 
 static void CPUMeter_display(Object* cast, RichString* out) {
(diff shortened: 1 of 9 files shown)


hishamhm:
Hi! Thanks for the PR!

This is a great idea. I have toyed around with how to display the CPU rate info but it didn't occur to me to simply replace the percentages. Not something to be on by default, but an interesting design option.

I'm concerned about having `clockRate` as a field in the generic `Meter` structure, though. I think this should be redesigned (possibly using a `Platform_getClockRate` function in `CPUMeter`). A new flag value like `Platform_hasClockRate` could also be used in Settings to hide that option in platforms that do not (yet) support it.

Explorer09:
Could I suggest that, instead of replacing the percentage (CPU time) in the CPUMeters, make clock rates a separate set of meters, with corresponding average?

shawnl:
I like this patch, but agree with the feedback, as it is important to hide this on not just platforms, but also architectures that do not support it.

`bash-4.4$ cat /proc/cpuinfo 
Processor	: ARMv7 Processor rev 2 (v7l)
processor	: 0
BogoMIPS	: 13.53

processor	: 1
BogoMIPS	: 13.53

processor	: 2
BogoMIPS	: 13.53

processor	: 3
BogoMIPS	: 13.53

Features	: swp half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt 
CPU implementer	: 0x51
CPU architecture: 7
CPU variant	: 0x0
CPU part	: 0x06f
CPU revision	: 2

Hardware	: QCT APQ8064 MAKO
Revision	: 000b
Serial		: 0000000000000000
`

Arnavion:
New PR since I would like this feature too: https://github.com/hishamhm/htop/pull/932

PR #596, opened 2017-02-10
Fix over-100% bug with old batteries

thukydides:
I had an issue with my old laptop battery.

Some batteries have less `CHARGE_FULL` than `CHARGE_FULL_DESIGN`,
but can still be charged to `CHARGE_FULL_DESIGN`,
meaning `CHARGE_NOW > CHARGE_FULL`.
This results in a battery level greater than 100%.

Checking `CHARGE_FULL_DESIGN` instead of `CHARGE_FULL` fixes this.
The same goes for `ENERGY_*`

See also [kernel documentation](https://www.kernel.org/doc/Documentation/power/power_supply_class.txt)

### DIFF
--- a/linux/Battery.c
+++ b/linux/Battery.c
@@ -234,7 +234,7 @@ static void Battery_getSysData(double* level, ACPresence* isOnAC) {
             if (!energy) {
                continue;
             }
-            const char* value = (!full) ? match(energy, "FULL=") : NULL;
+            const char* value = (!full) ? match(energy, "FULL_DESIGN=") : NULL;
             if (value) {
                totalFull += atoi(value);
                full = true;
--- a/linux/Battery.c
+++ b/linux/Battery.c
@@ -78,7 +78,7 @@ static unsigned long int parseBatInfo(const char *fileName, const unsigned short
       const unsigned long int foundNum = atoi(foundNumStr);
       free(foundNumStr);
       free(line);
-      
+
       total += foundNum;
    }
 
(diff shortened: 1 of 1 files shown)


Explorer09:
I don't like this change. CHARGE_FULL_DESIGN should never be considered by htop when drawing battery meter. Can we have a more precise percentage calculation that resemble the CAPACITY that the kernel reports? If it's above 100%, just top the value to 100%. No need to read two maximum values, right?

Explorer09:
If there's the case that CHARGE_NOW > CHARGE_FULL, that should a sign of a battery firmware bug that shouldn't be addressed in htop's side.

thukydides:
The kernel simply makes the firmware values available. So htop will have to handle something.
Also the kernel does not report any capacity attribute (Q&A at the bottom).

So level calculation will be in htop either way.
If we cap at 100% of `CHARGE_FULL`, the level may not be correct above that.
But I think you are right. Just capping at 100% is more reasonable.

You may still want to pull the cleanup commit 5b02fa587b7e5081f5dccccb8813f9f3569e4aa5.

thukydides:
I quickly did that. Available at thukydides@1dbed0d01754ea78c8d5272e0c7d6b4342b83e9f.

thukydides:
bump?

hishamhm:
> Just capping at 100% is more reasonable.

Yeah, I'll go with this simpler approach. Your commits had indentation issues, so it was quicker to rewrite and commit directly here. Thanks for bringing the issue to my attention!

PR #612, opened 2017-04-19
update keymap in help

ryenus:
I found `C` is used for setup, or configuration. meanwhile `F6` appears twice in the help page. So here's the fix.


hishamhm:
Thanks!

After I merged I remembered why F6 appears twice. It is actually used for both functions depending on context (whether one is in tree-view or not). Since that is pretty evident in the bottom function-keys row of the main UI, I decided to leave that as-is.

PR #613, opened 2017-04-19
Initial addition of dragonflybsd (based on FreeBSD)

dkgroot:
Small change in configure required to detect ncursesw.

Dragonflybsd specific solution
unique kernel process pid is not available. I had to fake it using kp_ktaddr which is unique. Process_writeField corrects this when writing to the screen.


hishamhm:
Yay! Thank you for the contribution! It is now merged. Looks like I need to publish a new release of htop soon. :)

dkgroot:
No problem at all.

Note: i have also create a deltaport for dragonflybsd in the meantime (Which include the same patch). Please give me a shout up when you publish a new release, so i can update the DeltaPort accordingly (removing the patch which would not be required anymore).

Thanks for htop BTW

etosan:
awesome!

PR #620, opened 2017-05-05
fix SEGFAULT when removing laptop battery

rindeal:
If I remove battery from my laptop and try to run htop, it segfaults, because the content of `/proc/acpi/battery/BAT0/info` changes, which results in `line` variable still containing NULL when `parseBatInfo()` passes it to `String_getToken()`. `String_getToken()` then supplies `strlen()` with this unchecked `line`.


hishamhm:
Merged, thank you! Sorry for taking so long!

PR #623, opened 2017-05-11
specify python version

aeTIos:
this will avoid compiling errors on distributions that use python3 as a global default (like, in my case, archlinux)


aeTIos:
I just noticed that this could also be fixed by changing the `import StringIO` line to `from io import StringIO` and keep the script version agnostic. Up to you.

hishamhm:
Does commit 50f03f59 fix this issue?

PR #624, opened 2017-05-11
make special keys also work with TERM=xterm-256color

mklein-de:


CRT.c
```diff
@@ -575,7 +575,7 @@ void CRT_init(int delay, int colorScheme) {
       CRT_scrollHAmount = 20;
    else
       CRT_scrollHAmount = 5;
-   if (String_eq(CRT_termType, "xterm") || String_eq(CRT_termType, "xterm-color") || String_eq(CRT_termType, "vt220")) {
+   if (String_eq(CRT_termType, "xterm") || String_startsWith(CRT_termType, "xterm-") || String_eq(CRT_termType, "vt220")) {
```
rindeal: why not just `String_startsWith(CRT_termType, "xterm")` ?

mklein-de: It's a bit less specific, but I guess it doesn't make a difference in practice.

PR #625, opened 2017-05-11
Added the github issues page link in the 'report a bug at' error message

bressan3:
Since the project is now on GitHub, I think it's better if we add its GitHub issues page address in the 'please report a bug' messages along with the official project page address.

### DIFF
--- a/darwin/DarwinCRT.c
+++ b/darwin/DarwinCRT.c
@@ -15,7 +15,8 @@ void CRT_handleSIGSEGV(int sgn) {
    (void) sgn;
    CRT_done();
    #ifdef __APPLE__
-   fprintf(stderr, "\n\nhtop " VERSION " aborting. Please report bug at http://hisham.hm/htop\n");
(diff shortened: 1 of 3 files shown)


Explorer09:
My personal opinion: If it's me, I would prefer not to have bug report URL hard-coded in .c files, but rather, use the macro PACKAGE_BUGREPORT and have the bug report address written in AC_INIT macro in configure.ac.

The reason for PACKAGE_BUGREPORT is that downstream distributors (e.g. distros) can change the URL easily if they wish to maintain their own htop branch.

hishamhm:
Hi.

Thanks for the suggestion! But I don't want to bundle a third-party URL in the source code. Last time when I moved from SourceForge to Github, I ended up with a ton of htop installations out there in the wild pointing at an outdated URL. That's why I changed it to a domain I own. If I need to move away from Github tomorrow, I don't want the same to happen again.

PR #636, opened 2017-06-14
wrap long command line with a separate window

ryenus:
This is quite useful for me to be able to inspect those processes with very long command lines to understand what command line options/arguments are used. I hope it can help others, too. Meanwhile C is definitely not my everyday programming language, so it would be great as well if someone could help improve this.

The feature being added is quite straightforward: select a process and press the <kbd>`L`</kbd> key, a new window is opened, and its command line, if couldn't fit into one line, would be wrapped properly around the window edge by the closest whitespaces.

This way it's much easier to see the whole command line at once, in the meantime we can use the builtin features to search and filter among the wrapped lines.

For example, since Google Chrome usually run its sub-processes with very long command lines, with this feature it would be much easier to know what command line flags were used.


ryenus:
A screenshot showing a Chrome command with 1.5K+ chars, for those interested in the result:

image

jtojnar:
This would be also useful for environment variables (#698)

PR #651, opened 2017-07-15
Round values in graph drawing (instead of implicit truncate)

Explorer09:
Hello.
I'm hoping to rework and adapt the "graph meter coloring & dynamic scaling" change (#487) so that it can merge on the current master. However, I wish to avoid repetitive "rebasing" and code tweaking in case of disagreements on some **minor behaviors**. Therefore I make this pull request.

This is more like a **question** than a pull to fix bugs.

The current behavior on the graph meter drawing is that values are **rounded toward zero** (a.k.a. truncate) to the number of dots (each bar) when drawing.
This proposal change to **round half away from zero**, using the round()/lround() function.

One of the result is that a CPU use of 99% may be presented as full bar on the graph, while in the current behavior, only strict 100.0% will be presented as full.

Will this be accepted, or should the current behavior be kept?

### DIFF
--- a/Meter.c
+++ b/Meter.c
@@ -406,8 +406,8 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
    }
    for (; i < nValues; i+=2, k++) {
       int pix = GraphMeterMode_pixPerRow * GRAPH_HEIGHT;
-      int v1 = CLAMP(data->values[i] * pix, 1, pix);
-      int v2 = CLAMP(data->values[i+1] * pix, 1, pix);
+      int v1 = CLAMP((int) lround(data->values[i] * pix), 1, pix);
+      int v2 = CLAMP((int) lround(data->values[i+1] * pix), 1, pix);
 
       int colorIdx = GRAPH_1;
       for (int line = 0; line < GRAPH_HEIGHT; line++) {


hishamhm:
> Will this be accepted, or should the current behavior be kept?

I can accept this behavior, as it reduces the overall rounding error (at the cost of an extra library call at a hot path, but improving visual correctness seems worth it).

However, I am still reluctant about #487. It adds a lot of code (both to the tool and, most importantly, for me to maintain), and the graph meter coloring adds a _lot_ of visual error,since the vertical resolution for colors is at most 4 (for example, causing sudden jumps when in fact there aren't). In this case, I'm gravitating towards thinking that it's better to not display information at all than to display very incorrect data.

Explorer09:
On Thu, Jul 27, 2017 at 3:15 AM, Hisham Muhammad <notifications@github.com>
wrote:
>
> I can accept this behavior, as it reduces the overall rounding error (at
> the cost of an extra library call at a hot path, but improving visual
> correctness seems worth it).

I think floating-point operations are already expensive enough that adding
a round operation shouldn't hurt too much on the performance.

> However, I am still reluctant about #487
> <https://github.com/hishamhm/htop/pull/487>. It adds a lot of code (both
> to the tool and, most importantly, for me to maintain),
>
That's one of the reasons I plan to rework. For me it's not about a lot of
code but because some of it should be separated into subroutines.

> and the graph meter coloring adds a *lot* of visual error,since the
> vertical resolution for colors is at most 4 (for example, causing sudden
> jumps when in fact there aren't).
>
The display can still be tuned if there are someone who worries about
sudden jumps. I welcome suggestions.

> In this case, I'm gravitating towards thinking that it's better to not
> display information at all than to display very incorrect data.
>
How about proposing the dynamic scaling change first? Then the coloring can
be added in another pull request.

hishamhm:
> I think floating-point operations are already expensive enough that adding a round operation shouldn't hurt too much on the performance.

Floating-point ops are pretty fast nowadays, but calling to a subroutine needs to adjust the stack, save registers, etc. I checked the asm generated by -O3 and was a bit surprised that `lround` was not replaced by a compiler intrinsic (but I don't want to add compiler/architecture specific code in there).

> That's one of the reasons I plan to rework. For me it's not about a lot of
code but because some of it should be separated into subroutines.

The amount of code still concerns me.

> The display can still be tuned if there are someone who worries about
sudden jumps. I welcome suggestions.

The trouble with your colored graph proposal is that the vertical "dot" resolution (16 braille dots, 6.25% each) and the color resolution (4 colors, 25% each) don't match (plus 25% is a bit too high for a reasonable graph IMO).

I don't think we have enough resolution on a Unix terminal to make a colored 4-char high vertical graph presenting reliable data. I mean, the only alternative I can think of is to make the graph blocky using U+2580 (▀) and U+2584 (▄) and hack foreground and background colors to fake an 8-char high display: then each block represents 12.5%, and the vertical and color resolutions match (which also simplifies the code), but then the horizontal resolution gets cut in half compared to braille.

_If_ I were to add that, I'd probably want to make it into a separate style, so at least the user can choose between color data or higher-res data.

> How about proposing the dynamic scaling change first? Then the coloring can
be added in another pull request.

I would like that, because dynamic scaling fixes an existing issue, which is some bogus graph visualizations the tool currently has.

Explorer09:
On Fri, Jul 28, 2017 at 1:08 AM, Hisham Muhammad
<notifications@github.com> wrote:
>
> Floating-point ops are pretty fast nowadays, but calling to a subroutine needs to adjust the stack, save registers, etc. I checked the asm generated by -O3 and was a bit surprised that lround was not replaced by a compiler intrinsic (but I don't want to add compiler/architecture specific code in there).

I read a little info on x86 floating point conversions just now. And
AFAIK even the simple float cast to (int) can bring the compiler to
emit a call to a helper routine (for example `_ftol()` for MSVC)
because of the hardware rounding mode adjustment.

So I think it is a tie when comparing with a lround() call. If there
are more optimization opportunities, they would be compilers' concern
and not ours.

hishamhm:
> If there are more optimization opportunities, they would be compilers' concern and not ours.

Agreed!

Issue #403, opened 2016-02-15
UI feature request: TAG column for process list


Explorer09:
A "tag" column can be added to the process list UI that shows check marks of whether a process is tagged (by space key) or not. This can help on the color scheme like monochrome or whenever it's not easy to distinguish tagged processes from untagged ones.
Also, with check boxes, it may be possible to "tag" processes by mouse click, [like such in Windows Explorer](http://www.howtogeek.com/howto/7888/how-to-enable-check-boxes-in-explorer-in-windows-7/), or "sort" the processes by whether they're tagged or not.
It can make some tasks convenient.

This is UI feature request.

PR #654, opened 2017-07-19
Make 'c' (tag process & its children) a toggle key

nckx:
Hullo,

I was a bit surprised that pressing ‘c’ is a one-way action: once a subtree is tagged, it appears that it can't be un-tagged as easily. This is my attempt to turn `tagAllChildren` into a more symmetrical toggle switch (while keeping things in relative sync):

- If the highlighted process was not tagged, current behaviour remains
  unchanged: that process and all its children are tagged.
- If it is, it is untagged along with all its children.

I.e. children aren't really toggled, but follow the parent. This allows tagging and untagging of subtrees in quite a natural fashion.

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -137,14 +137,12 @@ bool Action_setUserOnly(const char* userName, uid_t* userId) {
    return false;
 }
 
-static void tagAllChildren(Panel* panel, Process* parent) {
-   parent->tag = true;
+static void tagAllChildren(Panel* panel, Process* parent, bool value) {
+   parent->tag = value;
    pid_t ppid = parent->pid;
    for (int i = 0; i < Panel_size(panel); i++) {
       Process* p = (Process*) Panel_get(panel, i);
-      if (!p->tag && p->ppid == ppid) {
-         tagAllChildren(panel, p);
-      }
+      if (p->ppid == ppid) tagAllChildren(panel, p, value);
    }
 }
 
@@ -505,7 +503,7 @@ static Htop_Reaction actionUntagAll(State* st) {
 static Htop_Reaction actionTagAllChildren(State* st) {
    Process* p = (Process*) Panel_getSelected(st->panel);
    if (!p) return HTOP_OK;
-   tagAllChildren(st->panel, p);
+   tagAllChildren(st->panel, p, !p->tag);
    return HTOP_OK;
 }
 


nckx:
I left the help text unchanged, since \<space\> (which is also a toggle) is described in the same singular way.

nckx:
...I also just noticed ‘c’ is undocumented in the man page. Will add it tomorrow. :zzz:

Explorer09:
How about this?
'c' key untags the process and it's decedents only if all of itself and the decedents are tagged. I think this behavior will be more intuitive to users.

nckx:
Hm. This user would find that supremely unintuitive. :smile: 

More objectively, that behaviour would be predictable (and IMO useful) *only* on static process trees that don't spawn new children once tagged:

  **tagged_process** *hitting ‘c’ here will tag and untag both*
  └─ **tagged_child**

The new children aren't tagged (and probably shouldn't be?), making the result very dependent on context. Too unpredictable for my taste...

  **tagged_process** *hitting ‘c’ here **now** will do nothing*
  ├─ **tagged_child**
  └─ newly_spawned_untagged_child

The behaviour in this patch also won't break the way ‘c’ worked before (for tagging), keeping it backwards-compatible.

nckx:
> new children aren't tagged (and probably shouldn't be?)

I mean: it wouldn't cause any nasty surprises when killing, since you're killing the parent anyway, but it might when you're doing some fine-grained signal-sending and a new child pops up right before you hit the key.

Explorer09:
@nckx 
You should consider how htop kill processes with the 'k' key: It kill selected (tagged) processes one by one, not through a process group. And you can't prevent the race condition either way -- a process can spawn children milliseconds before it is delivered a kill signal.

For this reason, I'm for the idea that a process's new child should not be tagged when a tagged process has spawned it just now.

By the way, the child in the 'c' key is a misnomer, as what it tags are processes' _decedents_, not just its immediate children.

nckx:
@Explorer09: Yes, I agree?

hishamhm:
@nckx Hi, thanks for sending the patch!

I'm not a big user of the 'c' key myself, so I never noticed the annoyance. But yeah, making it a toggle sounds more useful. 

My only concern is the following scenario:

1. User tags a process with spacebar.
2. User then realizes they need to tag its children as well.
3. It goes back to the parent process and presses 'c'.

With the proposed patch, the result for the above scenario is that the parent process is untagged, with no visual feedback of what just happened.

nckx:
Hm. That *is* not good.

As noted above, I don't want to make this too context-dependent. But... maybe it would make sense to only treat a tagged process with no tagged sub-processes differently, and tag its descendants? Not sure.

Regarding the lack of visual feedback: I had noticed that there's no visual indication that a process is tagged when it's selected. I assumed there was good reason for this, in addition to bright yellow text on a bright cyan background being a sin against nature and/or certain terminals...

Ah, the joys of trivial patches :-)

hishamhm:
> Regarding the lack of visual feedback: I had noticed that there's no visual indication that a process is tagged when it's selected. I assumed there was good reason for this, in addition to bright yellow text on a bright cyan background being a sin against nature and/or certain terminals...

This was more of an oversight. But the color scheme can get complicated. Using "follow" (uppercase F) sets the background color of the selection bar yellow. So simply setting the foreground yellow wouldn't work.

I guess I could use to add a non-color visual indication for tagged processes:

```diff
-390,6 +390,9 @@ void Process_writeField(Process* this, RichString* str, ProcessField field) {
          baseattr = CRT_colors[PROCESS_THREAD_BASENAME];
       }
       if (!this->settings->treeView || this->indent == 0) {
+         if (this->tag) {
+            RichString_append(str, CRT_colors[PROCESS_TREE], "* ");
+         }
          Process_writeCommand(this, attr, baseattr, str);
          return;
       } else {
@@ -413,6 +416,9 @@ void Process_writeField(Process* this, RichString* str, ProcessField field) {
          const char* draw = CRT_treeStr[lastItem ? (this->settings->direction == 1 ? TREE_STR_BEND : TREE_STR_TEND) :
 TREE_STR_RTEE];
          snprintf(buf, n, "%s%s ", draw, this->showChildren ? CRT_treeStr[TREE_STR_SHUT] : CRT_treeStr[TREE_STR_OPEN]
 );
          RichString_append(str, CRT_colors[PROCESS_TREE], buffer);
+         if (this->tag) {
+            RichString_append(str, CRT_colors[PROCESS_TREE], "* ");
+         }
          Process_writeCommand(this, attr, baseattr, str);
          return;
       }
```

(...possibly with a nicer Unicode character.)

Explorer09:
On Fri, Jul 28, 2017 at 2:02 AM, Hisham Muhammad
<notifications@github.com> wrote:
>
> I guess I could use to add a non-color visual indication for tagged processes:
>
> -390,6 +390,9 @@ void Process_writeField(Process* this, RichString* str, ProcessField field) {
>           baseattr = CRT_colors[PROCESS_THREAD_BASENAME];
>        }
>        if (!this->settings->treeView || this->indent == 0) {
> +         if (this->tag) {
> +            RichString_append(str, CRT_colors[PROCESS_TREE], "* ");
> +         }
>           Process_writeCommand(this, attr, baseattr, str);
>           return;
>        } else {
> @@ -413,6 +416,9 @@ void Process_writeField(Process* this, RichString* str, ProcessField field) {
>           const char* draw = CRT_treeStr[lastItem ? (this->settings->direction == 1 ? TREE_STR_BEND : TREE_STR_TEND) :
>  TREE_STR_RTEE];
>           snprintf(buf, n, "%s%s ", draw, this->showChildren ? CRT_treeStr[TREE_STR_SHUT] : CRT_treeStr[TREE_STR_OPEN]
>  );
>           RichString_append(str, CRT_colors[PROCESS_TREE], buffer);
> +         if (this->tag) {
> +            RichString_append(str, CRT_colors[PROCESS_TREE], "* ");
> +         }
>           Process_writeCommand(this, attr, baseattr, str);
>           return;
>        }
>
> (...possibly with a nicer Unicode character.)

I'm bringing my feature request about tags (#403) here, I think mine
addresses similar problem.
Personally I don't like adding markers in the Command field as they
will break aligning on commands entries.

PR #656, opened 2017-07-23
more const usage

rsaxvc:
Mark some things as const, and gitignore some editor files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ htop
 # skip all backups
 *.bak
 *~
+.*.sw?
 
 # skip coverage files
 *.gcda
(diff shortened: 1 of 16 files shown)


hishamhm:
Thank you for contributing! :)

rsaxvc:
In files where they are not changed, it is a little cleaner and sometimes more memory-saving to declare them as const.

If, later, we want to update them in-place, as you suggest, all it takes is removing 'const'. But then you must also destroy and recreate the functionbar, since FunctionBar_new() caches the strings into RAM. Have you seen FunctionBar_setLabel()? That allows editing just one function's label.

PR #657, opened 2017-07-24
Settings: do not emit trailing whitespace

kalbasit:
I have `~/.config/htop` tracked under Git, and both Git and Vim are setup to trim trailing whitespace; after every htop use, git would display changes in the config file, whitespace changes.


hishamhm:
Trailing whitespaces don't annoy me (I actually use them to track indentation) and I don't usually add extra code to htop to deal with custom environments, but this is tiny enough to pass, and I understand a lot of people configure their Git that way and store their dotfiles in Git repos.  Thank you for contributing!

PR #660, opened 2017-07-27
call clear() function when SIGWINCH is received.

guoci:
redraw screen when SIGWINCH is sent to htop, even when terminal size remains unchanged.


hishamhm:
Hi! Thanks for contributing!

What problem does this solve? Can you describe a situation in which the current codebase produces wrong results that are fixed by this patch? Thank you!

guoci:
I am currently using SIGWINCH to redraw interactive programs when wall messages appear. This approach works on `top` and `watch` but not `htop`.

PR #664, opened 2017-08-01
tell GCC 7.x this is explicit fallthough

eworm-de:


hishamhm:
doesn't this break when compiled with clang (or other compilers)?

eworm-de:
At least `clang` does not break, but it reports its own warning now. *sigh*
```
ColumnsPanel.c:69:10: warning: declaration does not declare anything [-Wmissing-declarations]
         __attribute__((fallthrough));
```
Tested with `clang` version 4.0.1.

Explorer09:
Wait. Can't GCC detect the fallthrough keyword in comments?
Can we fix the comments as documented in `-Wimplicit-fallthrough` in [GCC Warning Options](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) page?

eworm-de:
I can't make any of these to work here, even when giving extra `-Wimplicit-fallthrough=1`. This is with gcc 7.1.1 on Arch Linux.

Explorer09:
Move the /* else fallthrough */ comment to outside the braces. Will that work?

eworm-de:
Ah, that's the issue! Looks like.

Let me do some more testing before I update the pull request.

eworm-de:
Updated...
Any comment on indention?

hishamhm:
@eworm-de Thank you!

Issue #665, opened 2017-08-02
Columns for delaystats


brendangregg:
Please add column for delaystats to htop. They've been in Linux for a long time, and show delays for different types of resources. There's not many tools that use them, so far.

The Linux kernel source ships with an example: https://github.com/torvalds/linux/blob/master/tools/accounting/getdelays.c

```
tools/accounting> ./getdelays -d -p $$
print delayacct stats ON
PID	4961


CPU             count     real total  virtual total    delay total  delay average
                  367      160000000      150819518        1942403          0.005ms
IO              count    delay total  delay average
                   40      306740944              7ms
SWAP            count    delay total  delay average
                    0              0              0ms
RECLAIM         count    delay total  delay average
                    0              0              0ms
```

It's the delay totals that are interesting. They are measured in nanoseconds, but presenting them in htop I think it would make more sense as percentages. Eg, to show 50% CPU delays for a process (thread) would mean it's spending half it's time waiting for a turn on CPU.

andrestc:
Is this a good starting issue for someone who is new to the project(does not know the codebase)? If so, I would like to try and tackle this.

brendangregg:
@andrestc it's a moderate change. A trivial addition would be a field that comes from a /proc file that htop is already reading. A bit harder would be adding a new /proc file that wasn't read before.

In this case, delaystats (sorry, I should be using the real name, which is "delay accounting") is read via the netlink interface, which is completely different, and more difficult, than a /proc reader. htop currently doesn't have a netlink reader (AFAIK). Fortunately, getdelays.c in the Linux source shows what needs to be done, but yes, it means adding a whole new data source to htop, which is why I'd call it moderate.

Is it good for a beginner? I'm not sure -- it would at least need a determined beginner who understood it's not, say, a 5 line trivial change. It'd probably end up a few hundred lines of code. Most of which may come from getdelays.c.

andrestc:
@brendangregg thanks for the detailed answer. 

I will give this a shot. But, as I might take slightly longer than someone with more experience on htop and its implementation details and I don't want to slow down anything, anyone can feel free to do it and beat me to it

hishamhm:
@andrestc Yes, that would be an excellent way to get started contributing to htop! @brendangregg's assessment is fully on point. It's not a quick patch, but it should be pretty self-contained, consisting basically of _additions_ rather than modifications to the code base.

[...]

PR #667, opened 2017-08-13
[WIP] Adds support for linux delay accounting

andrestc:
This PR adds support for showing columns with [linux delay accounting](https://www.kernel.org/doc/Documentation/accounting/delay-accounting.txt). This is a WIP but I wanted to open up for reviews to check if I'm on the correct track (cc @hishamhm).

This information can be read from the netlink interface, and thus we set up a socket to read from that when initializing the LinuxProcessList (`LinuxProcessList_initNetlinkSocket`). After that, for each process we call `LinuxProcessList_readDelayAcctData`, which sends a message thru the socket after setting up a callback to get the answer from the Kernel. That callback sets the process total delay time attribute. We then set the delay percent as the percentage of time process cpu time since last scan.

TODOS:
- [x] To compile it with delay accounting support I had to do:
`./autogen.sh && ./configure --enable-netlink CFLAGS=-I/usr/include/libnl3 && make`
I'm not sure there is a portable way to add that CFLAGS to the configure.ac file. libnl uses pkg-config for that, we could use [PKG_CHECK_MODULES](https://autotools.io/pkgconfig/pkg_check_modules.html) but I read that is not recommended. Figure out if there is a portable way to provide the CLFAGS required for libnl/libnl-gen
- [x] figure out why `./configure.sh` prints: `configure: WARNING: netlink/netlink.h: accepted by the compiler, rejected by the preprocessor!` (probably related to the item above)
- [x] add columns for other kinds of delays (e.g io)
- [x] add documentation on the new columns and decide on naming

Related to #665.

### DIFF
--- a/linux/LinuxProcess.c
+++ b/linux/LinuxProcess.c
@@ -82,9 +82,11 @@ typedef enum LinuxProcessFields {
    OOM = 114,
    IO_PRIORITY = 115,
    #ifdef HAVE_DELAYACCT
-   PERCENT_DELAY = 116,
+   PERCENT_CPU_DELAY = 116,
+   PERCENT_IO_DELAY = 117,
+   PERCENT_SWAP_DELAY = 118,
    #endif
-   LAST_PROCESSFIELD = 117,
+   LAST_PROCESSFIELD = 119,
 } LinuxProcessField;
 
 #include "IOPriority.h"
@@ -131,7 +133,11 @@ typedef struct LinuxProcess_ {
    #ifdef HAVE_DELAYACCT
    unsigned long long int delay_read_time;
    unsigned long long cpu_delay_total;
+   unsigned long long blkio_delay_total;
+   unsigned long long swapin_delay_total;
    float cpu_delay_percent;
+   float blkio_delay_percent;
+   float swapin_delay_percent;
    #endif
 } LinuxProcess;
 
@@ -224,7 +230,9 @@ ProcessFieldData Process_fields[] = {
    [OOM] = { .name = "OOM", .title = "    OOM ", .description = "OOM (Out-of-Memory) killer score", .flags = PROCESS_FLAG_LINUX_OOM, },
    [IO_PRIORITY] = { .name = "IO_PRIORITY", .title = "IO ", .description = "I/O priority", .flags = PROCESS_FLAG_LINUX_IOPRIO, },
 #ifdef HAVE_DELAYACCT
-   [PERCENT_DELAY] = { .name = "PERCENT_DELAY", .title = "DEL% ", .description = "Delay % accounting", .flags = 0, },
+   [PERCENT_CPU_DELAY] = { .name = "PERCENT_CPU_DELAY", .title = "CPUD% ", .description = "CPU delay %", .flags = 0, },
+   [PERCENT_IO_DELAY] = { .name = "PERCENT_IO_DELAY", .title = "IOD% ", .description = "Block I/O delay %", .flags = 0, },
+   [PERCENT_SWAP_DELAY] = { .name = "PERCENT_SWAP_DELAY", .title = "SWAPD% ", .description = "Swapin delay %", .flags = 0, },
 #endif
    [LAST_PROCESSFIELD] = { .name = "*** report bug! ***", .title = NULL, .description = NULL, .flags = 0, },
 };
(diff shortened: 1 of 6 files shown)


andrestc:
I've addressed my first two points in https://github.com/hishamhm/htop/pull/667/commits/94ef4e33512b3a333fdcfe834722b15f20dc451f. Ive also changed the feature flag to `enable-delayacct` to be more clear.
With the usage of `PKG_CHECK_MODULES` i was able to compile with:
`./autogen.sh && ./configure --enable-delayacct && make`

But we now require that the user compiling `htop` have `pkg-config` installed.

I forgot to mention on my PR, but we can only get the delay accounting running as root.

andrestc:
@hishamhm I the overall structure is done. I think I'll wait for your opinion before proceeding (for me it's only lacking the documentation and perhaps a better naming for the columns). Feel free to give your feedback when you have time.

hishamhm:
Hi @andrestc! Looking good!

There is some inconsistent indentation in some parts (reading the diff makes it easier to spot). Also, would it be possible to produce an error output in the column when `LinuxProcessList_readDelayAcctData` fails reading the data? (something like "N/A', like when the IO columns produce "no perm" when they have no permissions to read I/O info)

andrestc:
Hey @hishamhm. I think I got all the indentation right and also printed "N/A" when we get errors from the kernel.

andrestc:
Hi @hishamhm, just pinging to check if there is something I should update on the implementation. I'm waiting for the OK to get the documentation going.

Thanks!

hishamhm:
Ouch, I was misled by the Github interface: I had some comments written here that were marked as pending, and which I thought were published already. I'll go through them again!

linux/LinuxProcess.c
```diff
@@ -215,6 +229,11 @@ ProcessFieldData Process_fields[] = {
 #endif
    [OOM] = { .name = "OOM", .title = "    OOM ", .description = "OOM (Out-of-Memory) killer score", .flags = PROCESS_FLAG_LINUX_OOM, },
    [IO_PRIORITY] = { .name = "IO_PRIORITY", .title = "IO ", .description = "I/O priority", .flags = PROCESS_FLAG_LINUX_IOPRIO, },
+#ifdef HAVE_DELAYACCT
+   [PERCENT_CPU_DELAY] = { .name = "PERCENT_CPU_DELAY", .title = "CPUD% ", .description = "CPU delay %", .flags = 0, },
+   [PERCENT_IO_DELAY] = { .name = "PERCENT_IO_DELAY", .title = "IOD% ", .description = "Block I/O delay %", .flags = 0, },
+   [PERCENT_SWAP_DELAY] = { .name = "PERCENT_SWAP_DELAY", .title = "SWAPD% ", .description = "Swapin delay %", .flags = 0, },
+#endif
```
hishamhm: The lengths of the titles and the numeric displays (`%4.1f`) don't seem to match. I get the titles well aligned when I run it as a regular user and see "N/A" in all fields, but when I run as root and see the values, they are misaligned with the title column.

linux/LinuxProcessList.c
```diff
@@ -192,13 +206,30 @@ static void LinuxProcessList_initTtyDrivers(LinuxProcessList* this) {
    this->ttyDrivers = ttyDrivers;
 }
 
+#ifdef HAVE_DELAYACCT
+
+static void LinuxProcessList_initNetlinkSocket(LinuxProcessList* this) {
+  this->netlink_socket = nl_socket_alloc();
+  if (this->netlink_socket == NULL)
+    return;
```
hishamhm: Indentation is off here (also, I've been trying to add explicit brackets in if-statements in new code)

linux/LinuxProcessList.c
```diff
@@ -552,6 +589,69 @@ static void LinuxProcessList_readOomData(LinuxProcess* process, const char* dirn
    fclose(file);
 }
 
+#ifdef HAVE_DELAYACCT
```
hishamhm: Indentation in this block as well

linux/LinuxProcess.c
```diff
@@ -421,6 +454,14 @@ long LinuxProcess_compare(const void* v1, const void* v2) {
    #endif
    case OOM:
       return (p2->oom - p1->oom);
+   #ifdef HAVE_DELAYACCT
```
hishamhm: These comparisons are not working because they are float values. See the PERCENT_CPU comparison in Process.c.

andrestc:
Thanks for the review, @hishamhm. I think i've addressed all your comments.

hishamhm:
Awesome, thanks! I think the column names are okay, so the only thing left in your TODO list is documentation; additions to the manpage should suffice.

andrestc:
Manpage updated with the new columns. PTAL, @hishamhm.

hishamhm:
Great, thank you! Merging!

ygoe:
In what htop version will I find this? There are no meaningful release tags here on GitHub.

PR #674, opened 2017-08-24
RFC: Experimental support for metering zswap

daniel-thompson:
Liunx' zswap feature works by allocating pages from the "real" swap but
delaying the process of writing them out, leaving them compressed in RAM
instead. It therefore has three interesting values for metering:

 * max size for compressed pages (max_pool_size)
 * actual size of compressed pages in memory (pool_total_size)
 * quantity of data not yet written to real swap (stored_pages)

This patch graphs stored_pages are part of the existing swap meter
because the deferred write-out means that stored_pages always represents
a subset of the total number of swapped out pages. It also introduces a
new zswap meter (hidden by default) to show how much of the maximum
compressed space is currently in use.

This patch is an RFC because it depends upon (potentially unstable)
debugfs attributes that can only be read when running as root. The
main purpose of the RFC is to guage interest in proposing a less
root-centric ABI for zswap. Put another way, if this patch were able
to use real sysfs attributes would it be acceptable for htop?


daniel-thompson:
I'll treat no comments after six months and an clear answer here!

PR #676, opened 2017-09-01
Find roots when constructing process tree, fix #587

### LINKED ISSUE #587: Ordering bug in tree view
On my system, processes by other users aren't visible in htop due to SELinux restrictions. This includes the init process. Normally, this doesn't bother htop, but it seems to cause the following bug (which does not occur:

When a process has a lower PID than its parent, it is not included in the tree below its parent; instead, it is listed above the parent without any indentation. Here's a screenshot: 

htop_tree_order_bug

This does not happen when run htop as root (thereby overriding the restrictions mentioned above); hence I assume this bug is causes by htop not being able to see the whole process tree.

wangqr:
I'm maintaining a multi-user server with /proc mounted with option `hidepid=2`, so each user can only see his own processes. Users with long running tmux sessions will see some of their processes not correctly shown under its parent when using tree view in htop. So it is difficult to find all processes running inside tmux.

After some search I found #587 which is related to this issue. This PR fix the issue by only choosing those processes as tree roots whose parents are not seen in the process list.

Here is a comparison: (Left: current master, Right: this PR, see PID 4157)
default

### DIFF
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -213,23 +213,45 @@ void ProcessList_sort(ProcessList* this) {
       // Restore settings
       this->settings->sortKey = sortKey;
       this->settings->direction = direction;
-      // Take PID 1 as root and add to the new listing
       int vsize = Vector_size(this->processes);
-      Process* init = (Process*) (Vector_take(this->processes, 0));
-      if (!init) return;
-      // This assertion crashes on hardened kernels.
-      // I wonder how well tree view works on those systems.
-      // assert(init->pid == 1);
-      init->indent = 0;
-      Vector_add(this->processes2, init);
-      // Recursively empty list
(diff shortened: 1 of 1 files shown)


hishamhm:
Hi! Thank you for contributing!

hishamhm:
Merged it now. I'll add a test so that it only runs the binary search when the current pid is not 1. Pid 1 is always a root, and in most systems all processes will be visible, so it will always be the first entry, so we can skip the bisection, go straight to building the tree, and it will be node in one round.

wangqr:
`init` will have PPID=0, which is the kernel scheduler and should not be in the process list, so search for it won't find anything. If doing a special test for `init` is for performance considerations, using the following condition might be faster:
* test if PPID=0, or
* test if PPID < PID

PR #457, opened 2016-03-21
Clean up trailing whitespaces

Explorer09:
Trailing whitespaces (and whitespace indents in otherwise blank lines)
are bringing me unnecessary merge conflicts when I try merging local
branches. It's good to clean up all of them at once.

(And yes I'll rebase my other pull requests here upon this one.)

PR #677, opened 2017-09-10
Fix build with glibc 2.25

Bfgeshka:
More info here: https://sourceware.org/ml/libc-alpha/2017-02/msg00079.html

Removing trailing spaces as a bonus from my editor, didn't notice it when commited.

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -16,6 +16,7 @@ in the source distribution for its full text.
 #include <stdio.h>
 #include <sys/time.h>
 #include <sys/resource.h>
+#include <sys/sysmacros.h>
 #include <sys/param.h>
 #include <sys/stat.h>
 #include <sys/types.h>
@@ -204,7 +205,7 @@ void Process_setupColumnWidths() {
 void Process_humanNumber(RichString* str, unsigned long number, bool coloring) {
    char buffer[11];
    int len;
-   
+
    int largeNumberColor = CRT_colors[LARGE_NUMBER];
    int processMegabytesColor = CRT_colors[PROCESS_MEGABYTES];
    int processColor = CRT_colors[PROCESS];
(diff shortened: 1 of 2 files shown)


Explorer09:
Can you explain exactly what did you fix? How does the new `#include <sys/sysmacros.h>` matter in the build?

Bfgeshka:
It is explained in release notes I've attached to PR:

 
>  The inclusion of <sys/sysmacros.h> by <sys/types.h> is deprecated.
>  This means that in a future release, the macros “major”, “minor”, and
>  “makedev” will only be available from <sys/sysmacros.h>.
>
>  These macros are not part of POSIX nor XSI, and their names frequently
>  collide with user code; see for instance glibc bug 19239 and Red Hat
>  bug 130601.  <stdlib.h> includes <sys/types.h> under _GNU_SOURCE, and
>  C++ code presently cannot avoid being compiled under _GNU_SOURCE,
>  exacerbating the problem.

Explorer09:
@Bfgeshka 
The problem is, your supposed "fix" breaks the build with clang. Did you read the Travis CI build result?
With clang it says:
```
clang -DHAVE_CONFIG_H -I.  -DNDEBUG  -pedantic -Wall -Wextra -std=c99 -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR=\"/usr/local/etc\" -I"./darwin"  -g -O2 -MT htop-Process.o -MD -MP -MF .deps/htop-Process.Tpo -c -o htop-Process.o `test -f 'Process.c' || echo './'`Process.c
Process.c:19:10: fatal error: 'sys/sysmacros.h' file not found
#include <sys/sysmacros.h>
         ^
1 error generated.
```
So maybe you should use the configure script to check the existence of <sys/sysmacros.h> before including it. <sys/sysmacros.h> is not a standard header, hence you can't assume it exists in every system.

Also, I don't think @hishamhm will agree to have removal of trailing whitespaces in this pull request (I proposed that before, #457 ), so let's just keep your patch small.

Bfgeshka:
Nah, I didn't really read it before.

Well, I guess someone more interested will supply better patch, I'm leaving this one as my local application then since I have no will to deal with autotools at all.

PR #678, opened 2017-09-22
Remove libtool references

evelikov:
The project builds a single standalone binary.

There are no libraries created - be that static or shared ones.
Thus there's no need for libtool.


evelikov:
Humble ping anyone?

hishamhm:
Indeed. Thank you!

PR #679, opened 2017-09-22
Hide internal symbols from the final binary

evelikov:
Probe for -fvisibility=hidden and use it.
This way all the internal symbols are hidden, reducing the size of the
binary by ~10%.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,7 +12,7 @@ applications_DATA = htop.desktop
 pixmapdir = $(datadir)/pixmaps
 pixmap_DATA = htop.png
 
-htop_CFLAGS = -pedantic -Wall $(wextra_flag) -std=c99 -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR=\"$(sysconfdir)\" -I"$(top_srcdir)/$(my_htop_platform)"
+htop_CFLAGS = -pedantic -Wall $(wextra_flag) $(fvisibility_flag) -std=c99 -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR=\"$(sysconfdir)\" -I"$(top_srcdir)/$(my_htop_platform)"
 htop_LDFLAGS = 
 AM_CPPFLAGS = -DNDEBUG
 
--- a/configure.ac
+++ b/configure.ac
@@ -100,6 +100,19 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([], [])],[
 CFLAGS="$save_cflags"
 AC_SUBST(wextra_flag)
 
+save_cflags="${CFLAGS}"
+CFLAGS="$CFLAGS -fvisibility=hidden"
+AC_MSG_CHECKING([if compiler supports -fvisibility=hidden])
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([], [])],[
+   fvisibility_flag=-fvisibility=hidden
+   AC_MSG_RESULT([yes])
+],[
+   fvisibility_flag=
+   AC_MSG_RESULT([no])
+])
+CFLAGS="$save_cflags"
+AC_SUBST(fvisibility_flag)
+
 # Checks for features and flags.
 # ----------------------------------------------------------------------
 PROCDIR=/proc


Explorer09:
Is reducing file size your only reason for proposing to use this compiler option? Htop is no library and it has no C++ code, so there're not many benefits for using `--fvisibility`. I wonder what prevents you from using `strip` to reduce the file size?

evelikov:
Stripping (A) the binary is orthogonal to hiding internal symbols (B).

Some things that come to mind:
A) Privacy reasons (say binary-only module), reduce the size or split out debug symbols into separate file.
B) Prevent symbol collision, reduce the binary size and overall be a good citizen.

Explorer09:
There are no symbols "internal" enough to hide for htop, unfortunately. Like I said, htop has no library code and no C++, what it is to hide, anyway?

Symbols collisions that -fvisibility addresses is when linking with multiple shared libraries. It doesn't solve collisions when linking .o files (that's what "static" was for).

And for size reduction, can you show me the two versions of the binary, with and without the switch? Also it's better to show whether they reduce to same binary after the strip command.

evelikov:
There's no 'enough' in '"internal" enough' ;-) They are either internal or not.

There are executables that intentionally export API, that is used by plugins.
htop does not have any, so don't expose the symbols..

Some numbers, as requested. Note that those will differ depending on the configure toggles, compiler version flags, etc. Feel free to apply the patch and check locally.

- w/o -fvisibility
855K htop
178K htop.stripped

- with -fvisibility
839K htop
162K htop.stripped

Explorer09:
And how about the comparison of symbol tables?

Explorer09:
Ok, so I tried building htop myself with and without `-fvisibility=hidden`, and I got a brief idea of what's going on. The approximate 16kiB reduction is not due to .text section reductions, but due to removal of several symbols from dynsym table (plus the corresponding `.gnu.hash`, `.dynstr` and `.gnu.version` sections). 

But still, I'm not considering -fvisibility=hidden the right way to deal with the problem. Is there a way to change the linking behavior to not export these symbols to dynsym in the first place?

evelikov:
What goes it which section (modulo a few generic rules) is compiler specific.
One way to get .text reduction [with the toggle] is to have dead code in htop.

Using `-fvisibility=hidden` is the way to tell the compiler - hey these are internal, do your magic. @Explorer09 I don't see any technical arguments so far nor alternative solution provided. I'm starting to feel that you like bikeshedding  ;-(

In case you're wondering - why does the compiler not use the toggle (or equivalent) by default, say for executables at least? Because it will break apps which intentionally expose ABI for their plugins - bash, zsh, irssi, vim, Xorg (and friends), Weston ...

Explorer09:
@evelikov I have already told you that the size reduction isn't from .text reduction at all. I've compared the disassembly. The generated code of every function in .text section is identical except pointer addresses' changes.

And mind you, executable programs (.exe or no extension) are never meant for further linking by other programs. That's what .so and .dll files are for. So it's considered an error to leave symbols in .dynsym sections except library calls. This error only happens in linking phase of htop program, for now.

evelikov:
> @evelikov I have already told you that the size reduction isn't from .text reduction at all.

Where did I claim the opposite?

> And mind you, executable programs (.exe or no extension) are never meant for further linking by other programs. That's what .so and .dll files are for. So it's considered an error to leave symbols in .dynsym sections except library calls. This error only happens in linking phase of htop program, for now.

As mentioned before - some applications require the symbols to be exported (bash, zsh....), hence they should be in .dynsym. htop has no plugins that use the API so lets hide it?

@hishamhm can you weight it? I fear that we're going in circles with @Explorer09 :-(

Explorer09:
@evelikov The problem isn't about hiding, it's about why they should be exported (to dynsym) when they're never used outside the program binary.

I think I found the culprit: The `-rdynamic` flag (look in Makefile.am) where htop is still used when building for Linux.
Why the flag is still there and not yet removed?

evelikov:
`-rdynamic` implies `--export-dynamic`, aka "export all". So apart from the htop local symbols, some extra ones are exposed as well (from a few static objects implicitly used at linktime).

`__data_start
data_start
_IO_stdin_used
__libc_csu_fini
__libc_csu_init
_start`

Allegedly -rdynamic is useful for debugging/backtraces/etc and tools like cmake even always use it.
Although dropping it will resolve the "issue" as well.

Explorer09:
Your proposed `-fvisibility=hidden` will hinder backtraces just as removing `-rdynamic` will.

The best you can propose to htop is a `--disable-backtrace` configure option, which will disable htop's backtrace as well as `-rdynamic` and make binary smaller.

I think we've resolved the debate.

evelikov:
For the future, @Explorer09 please read more carefully what people say. Admitting when you get something wrong, is a big plus.

I have not seen other applications have a such toggle. So it seems there's a 'debate' just for the sake of it. Proceed as you wish - I had enough of this bike-shedding.

This is one the worst Github PR so far - "I don't like it, but it will take me 4-5 replies explain why or suggest alternatives"

Explorer09:
@evelikov If you have explained more about your reasoning in the first place. I did read carefully, but your only initial argument for this PR is size, which is not a good reason considering other trade-offs.
It was not until the last couple of your posts did I read that you didn't care about backtrace (and crash-reporting) at all. Even then, `-fvisibility=hidden` is wrong fix.

evelikov:
> @evelikov If you have explained more about your reasoning in the first place.

Correct. My first reply covers all that was missing.

> I did read carefully,

Not enough it seems. You misread the .text part, plus the multiple reminders that some executables must export internal symbols got completely ignored.

> but your only initial argument for this PR is size, which is not a good reason considering other trade-offs.

Where and more importantly _when_ were those trade-offs mentioned?

> It was not until the last couple of your posts did I read that you didn't care about backtrace (and crash-reporting) at all. 

I never said or implied that. Not sure what gave you such a crazy idea.

Good luck!

Explorer09:
> You misread the .text part, plus the multiple reminders that some executables must export internal symbols got completely ignored.

The reason for exporting in other software is irrelevant to htop's case. If you have not told about backtrace later, I would consider your argument making no sense. That's why I ignored that.

> Where and more importantly when were those trade-offs mentioned?

It's generally a bad idea to strip off something for size unless you really know what you were doing. I have been compiling programs, too, and I knew many ways to bring down binary sizes, each with its own drawbacks. This should be common sense, just like you are taught not to use `-O3` to compile everything  unless you know it's safe (or when the project assumes `-O3` by default). If I am aiming for size, I would go at least `-Os -flto -fno-asynchronous-unwind-tables`. For other flags, I have an instinct to feel _something might go wrong_ whenever anything is dropped from the binary.

> Not sure what gave you such a crazy idea.

People report htop crashes with backtraces always. Hiding the symbols as you suggested will simply make it harder to make a proper report. That's the trade-off for the size. So I cannot find a good reason for your `-fvisibility` proposal other than you don't care about backtrace and crash reporting.

hishamhm:
Closing this one since the symbols are valuable for crash backtraces.

evelikov:
Food for thought:
Did anyone check a backtrace with the proposed patch? It seemed perfectly legible here, so either there's something special on my end or this was a mere bikeshedding exercise :-\

Explorer09:
@evelikov "Legible" is a subjective adjective. Better show us a comparison of backtraces between with and without the patch. However, I personally won't bother to do it.

PR #689, opened 2017-10-07
Minor quoting fix in configure.ac

Explorer09:
This lack of quotes breaks syntax highlighting of gedit.
No effect on configure script generation, though.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -268,7 +268,7 @@ if test ! -z "$missing_headers"; then
   AC_MSG_ERROR([missing headers: $missing_headers])
 fi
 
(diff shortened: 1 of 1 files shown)


hishamhm:
@Explorer09 shouldn't you send a PR to gedit instead? Sounds like a bug in their highlighting, not in htop.

Explorer09:
@hishamhm
I'll need to test whether it happens in latest gedit version, since mine is quite old (3.10.4 from Ubuntu 14.04). Besides, it _will_ fail to parse when the copyright string mentioned has unpaired parentheses, or a comma. Adding a pair of M4 quotes there can help avoid the problem anyway.

Explorer09:
Reported: https://bugzilla.gnome.org/show_bug.cgi?id=788973

Issue #180, opened 2015-04-06
starttime resets if /proc remounted


shawnl:
https://github.com/shawnl/htop/commit/edc4ea90f20986eb78af7de39497a6de693209bb
https://github.com/shawnl/htop/commit/ffd304321735c20a8a9fb004f48a29d7ffd30a03

hishamhm:
Can't this be calculated once when the process is added instead of every time the process is displayed?

shawnl:
Yes it could. But the information should come from the stat file, cause thats the best source.

shawnl:
I don't think it makes much of a differnce if this gets updated every time, considering that this file is already being read.

joder:
I am bitten by this issue too. Will the patch by @shawnl ever be merged or would the issue better be resolved in another way?

hishamhm:
My question about his patch was about triggering an extra syscall on each list refresh (getting `localtime_r`) in Process.c, but I haven't tested it yet; it's possible that the glibc caches this information. The current codebase builds the starttime string once and just reuses that. I like the idea of reading from the stat file, but I don't want a performance penalty. I will revisit this issue when I get back at the code, thank you for your feedback.

shawnl:
localtime_r() is not a real syscall, it gets handled by the VDSO http://man7.org/linux/man-pages/man7/vdso.7.html

sh3bang:
Bug not fixed? STARTIME is still wrong!

shawnl:
Not if you remount /proc

sh3bang:
i dount mount or remount /proc... its done during boot. But htop show wrong start time... instead of top

shawnl:
This was never merged

hishamhm:
@shawnl Thanks for applying the changes to the PR. I will give it a spin an merge it for the upcoming release!

folfy:
Just stumbled upon this problem while using htop (very odd behavior indeed).
@hishamhm Anything we can help/contribute to the PR? Is there something left which is preventing this from being merged?

hishamhm:
Merged using @shawnl's fix. Thank you!

PR #700, opened 2017-11-22
get STARTTIME from proc

shawnl:
see bug https://github.com/hishamhm/htop/issues/180

### DIFF
--- a/linux/LinuxProcess.c
+++ b/linux/LinuxProcess.c
@@ -15,6 +15,7 @@ in the source distribution for its full text.
 #include <unistd.h>
 #include <string.h>
 #include <sys/syscall.h>
+#include <time.h>
 
 /*{
 
@@ -100,6 +101,7 @@ typedef struct LinuxProcess_ {
    long m_drs;
    long m_lrs;
    long m_dt;
+   unsigned long long starttime;
    #ifdef HAVE_TASKSTATS
    unsigned long long io_rchar;
    unsigned long long io_wchar;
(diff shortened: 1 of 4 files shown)


shawnl:
In my testing glibc caches the access to /etc/localtime

shawnl:
I don't have a mac so I can't test on that platform.

Process.c
```diff
@@ -443,7 +446,17 @@ void Process_writeField(Process* this, RichString* str, ProcessField field) {
    }
    case PROCESSOR: xSnprintf(buffer, n, "%3d ", Settings_cpuId(this->settings, this->processor)); break;
    case SESSION: xSnprintf(buffer, n, Process_pidFormat, this->session); break;
-   case STARTTIME: xSnprintf(buffer, n, "%s", this->starttime_show); break;
+   case STARTTIME: {
+#ifdef __linux__
```
hishamhm: This should be moved to `linux/LinuxProcess.c`, where it will override the one from `Process.c`.

In general, code at the root directory files should be portable, and not have any `ifdef __linux__` in them.

htop.c
```diff
@@ -162,6 +167,24 @@ static void millisleep(unsigned long millisec) {
    }
 }
 
+static void record_btime() {
```
hishamhm: This operation can be done in `linux/LinuxProcessList.c`, in the initialization function of that file.

shawnl:
thanks for the review! Much better now.

hishamhm:
Merged, thank you! Sorry about the delay.

PR #702, opened 2017-11-24
Clarify we are looking for the null termination

coypoop:
Not for a comparison to zero


hishamhm:
Thank you! Representing intention in code is always important :)

PR #708, opened 2017-12-02
Run modern gcc

szepeviktor:
gcc-6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,16 @@ os:
   - linux
   - osx
 
+matrix:
+  include:
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-6
+      env:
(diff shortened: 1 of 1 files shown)


szepeviktor:
There is one warning only: https://travis-ci.org/hishamhm/htop/jobs/310520033#L837

hishamhm:
Hi! Thanks for looking into this! However, the description mentions gcc, but that build ended up running clang. So maybe there's a problem in this patch?

szepeviktor:
@hishamhm The bottom one is the new gcc build: https://travis-ci.org/hishamhm/htop/builds/310521395

hishamhm:
@szepeviktor from the looks of the Travis logs, it seems this is still running gcc 4: https://travis-ci.org/hishamhm/htop/jobs/310521400

szepeviktor:
@hishamhm It may be because of a temporary network failure: https://travis-ci.org/hishamhm/htop/jobs/310521400#L422
Package manager was not able to install gcc 6.
Could you restart the build?

PR #712, opened 2018-01-02
Fix infinite loop in tree view on Mac OS X

### LINKED ISSUE #688: display nothing on mac os 10.13 lastest version
display nothing.

wataash:
Quote from the commit log:

> Fixes #688, the bug regressed on 584a9bc.
> 
> On Mac OS X 10.11.6, all processes have their parents since there's a
> special process named "kernel_task", whose PID and PPID are 0. As a
> result, `this->processes` is never changed causing infinite `while`.

Tested on Mac OS X 10.11.6 and Ubuntu 16.04.

### DIFF
--- a/ProcessList.c
+++ b/ProcessList.c
@@ -231,6 +231,11 @@ void ProcessList_sort(ProcessList* this) {
             pid_t ppid = process->tgid == process->pid ? process->ppid : process->tgid;
             // Bisect the process vector to find parent
             int l = 0, r = size;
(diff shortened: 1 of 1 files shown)


wataash:
It doesn't occur on macOS Sierra because "kernel_task" doesn't apper... sorry :(
I'll rewrite all as s/macOS/Mac OS X/g .

hishamhm:
Thank you, @wataash!

Issue #682, opened 2017-09-26
Htop on macOS High Sierra


MenkeTechnologies:
Can anyone confirm the htop available on homebrew (2.02) is broken on macOS 10.13 (crashes iTerm2 after several seconds)?

fykuan:
I am facing a similar issue. After several seconds running htop in iTerm2 (and Terminal.app too). The terminal halts and I have to force close it. After that, iTerm2 won't open again and macOS refuses to reboot. I have to force reboot macOS by holding the power button.

anothernode:
I'm experiencing the exact same behavior as @fykuan describes with the only difference that htop can run for quite a while before it crashes on my machine, I'd estimate up to half an hour or so.

I just asked on [apple.stackexchange.com](https://apple.stackexchange.com/questions/300185/is-anyone-else-having-problems-with-using-htop-on-high-sierra) about this and at least one person replied in the comments reporting the same problem.

I installed htop (version 2.0.2) with `brew install htop`.

Nax:
I experience the same problem. Like @anothernode, it happens after a few hours on my machines.
Htop version 2.0.2, installed from MacPorts.

amertkara:
Having the same issue. iTerm runs fine on its own until I start htop. Initially htop runs fine but if I keep it running for a while (more than a couple of minutes), I see the spinning pointer and eventually have to force quit iTerm. I cannot start both iTerm and Terminal apps without restarting the laptop.

cusster:
Good to know I'm not the only one having this issue. Using iTerm2 and htop is installed using homebrew. Hopefully this gets fixed soon.

shevegen:
> After that, iTerm2 won't open again and macOS refuses to reboot.

What the ...

I am glad to say that htop for me, once compiled, always worked on Linux - I never had a situation where Linux would fail to reboot in such a situation ... :)

I guess for the time being one should not use htop on High Sierra until it is fixed; wonder what changed if it worked before ...

spprichard:
I'm experiencing the same issue htop: 2.0.2 MacOS: 10.13 Build(17A365). Anyone know a fix?

roopej:
I can also confirm the same on macOS 10.13 with htop 2.0.2 running in iTerm2 build 3.0.3. After some time I get the spinning pointer. Activity Monitor goes totally blank and MBP needs to be rebooted.

nayzak:
[...]

PR #713, opened 2018-01-03
Add notes about running on macOS without sudo

wataash:


wataash:
I close it bacause giving htop setuid flag has security risks as @ryenus and @dos1 mentioned in #682.

PR #716, opened 2018-01-14
Add support for Linux TASK_IDLE

CyberShadow:
```
Linux commit 06eb61844d841d0032a9950ce7f8e783ee49c0d0 ("sched/debug:
Add explicit TASK_IDLE printing") exposes kthreads idling using
TASK_IDLE in procfs as "I (idle)".

Until now, when sorting the STATE ("S") column, htop used the raw
value of the state character for comparison, however that led to the
undesirable effect of TASK_IDLE ('I') tasks being sorted above tasks
that were running ('R').

Thus, explicitly recognize the idle process state, and sort it below
others.
```

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -174,6 +174,8 @@ typedef struct ProcessClass_ {
 
 #define Process_isChildOf(process_, pid_) (process_->tgid == pid_ || (process_->tgid == process_->pid && process_->ppid == pid_))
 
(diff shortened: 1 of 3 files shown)


hishamhm:
Merged, thank you!

CyberShadow:
Thanks!

Not sure you were aware, but the referenced Linux commit mentions htop explicitly in the commit message. :)

PR #719, opened 2018-01-23
Put an upper bound on the number of runningTasks reported

lucasvr:
htop scans /proc to determine how many running tasks exist. Since that operation is not possible to be conducted in an atomic fashion, task scheduling effects can lead to a count greater than the number of actual processors; this is more easily noticed on machines with several CPUs and under heavy workload.

This patch simply adds an upper bound on cpuCount to guarantee consistent reports of the number of running tasks at any given time.


hishamhm:
Thank you!

PR #721, opened 2018-01-24
Allow interactive command to pause updates

wrp:
This is more a request for a feature implementation than anything else.  This patch works for me, but there's probably a better implementation.

### DIFF
--- a/ScreenManager.c
+++ b/ScreenManager.c
@@ -191,6 +191,7 @@ void ScreenManager_run(ScreenManager* this, Panel** lastFocus, int* lastKey) {
       int prevCh = ch;
       set_escdelay(25);
       ch = getch();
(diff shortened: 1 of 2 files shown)


szepeviktor:
When I kill a process it *goes away*

hishamhm:
@szepeviktor yes, this is not as easy as this commit suggests, because the way this commit does it would affect all ScreenManagers and not just the MainPanel. One way to do it would be as an Action event that sets a toggle flag which then causes checkRecalculation to be skipped.

I'm not sure about what's the best behavior for pausing though. I think some people would like pause to stop and wait for a key, like this patch, others would like pause to pause updates but have the screen remain scrollable (as the suggestion I made would do).

PR #728, opened 2018-02-04
keep scan thread for version before high sierra

pmalhaire:
@hishamhm As said this should do the trick to keep thread info for those who don't have High Sierra.

Unfortunately I don't know the exact version failling.

### DIFF
--- a/darwin/DarwinProcessList.c
+++ b/darwin/DarwinProcessList.c
@@ -18,6 +18,53 @@ in the source distribution for its full text.
 #include <sys/mman.h>
 #include <utmpx.h>
 #include <err.h>
+#include <sys/sysctl.h>
+#include <stdbool.h>
+
+void GetKernelVersion(int * major, int * minor, int * internal) {
+   static int cMajor = 0;
+   static int cMinor = 0;
+   static int cInternal = 0;
+   if (!cMajor) {
+      // just in case it fails someday
+      cMajor = cMinor = cInternal = -1;
+      char str[256] = {0};
+      size_t size = sizeof(str);
+      int ret = sysctlbyname("kern.osrelease", str, &size, NULL, 0);
+      if (ret == 0) {
+         sscanf(str, "%d.%d.%d", &cMajor, &cMinor, &cInternal);
+      }
+    }
+    *major = cMajor;
+    *minor = cMinor;
+    *internal = cInternal;
+}
+
+bool IsKernelVersionUpperOrEqual(int major, int minor, int internal) {
+   int maj, min, inter;
+
+   GetKernelVersion(&maj, &min, &inter);
+
+   if ( maj > major) return true;
+   if ( maj == major) {
+      if ( min > minor ) return true;
+      else if ( min == minor && inter >= internal ) return true;
+   }
+   return false;
+}
+
+bool IsKernelVersionLessOrEqual(int major, int minor, int internal) {
+   int maj, min, inter;
+
+   GetKernelVersion(&maj, &min, &inter);
+
+   if ( maj < major) return true;
+   if ( maj == major) {
+      if ( min < minor ) return true;
+      else if ( min == minor && inter <= internal ) return true;
+   }
+   return false;
+}
 
 /*{
 #include "ProcessList.h"
@@ -170,9 +217,13 @@ void ProcessList_goThroughEntries(ProcessList* super) {
 
        DarwinProcess_setFromKInfoProc(&proc->super, &ps[i], tv.tv_sec, preExisting);
        DarwinProcess_setFromLibprocPidinfo(proc, dpl);
-       
-       // Disabled due to bug in macOS High Sierra
-       // DarwinProcess_scanThreads(proc);
+
+       // Disabled for High Sierra due to bug in macOS High Sierra
+       bool isScanThreadSupported  = ! ( IsKernelVersionUpperOrEqual(17, 0, 0) && IsKernelVersionLessOrEqual(17, 4, 0));
+
+       if (isScanThreadSupported){
+           DarwinProcess_scanThreads(proc);
+       }
 
        super->totalTasks += 1;
 
--- a/darwin/DarwinProcessList.c
+++ b/darwin/DarwinProcessList.c
@@ -21,49 +21,30 @@ in the source distribution for its full text.
 #include <sys/sysctl.h>
 #include <stdbool.h>
 
-void GetKernelVersion(int * major, int * minor, int * internal) {
-   static int cMajor = 0;
-   static int cMinor = 0;
-   static int cInternal = 0;
-   if (!cMajor) {
+struct kern {
+    short int version[3];
+};
+
+void GetKernelVersion(struct kern *k) {
+   static short int version_[3] = {0};
+   if (!version_[0]) {
       // just in case it fails someday
-      cMajor = cMinor = cInternal = -1;
+      version_[0] = version_[1] = version_[2] = -1;
       char str[256] = {0};
       size_t size = sizeof(str);
       int ret = sysctlbyname("kern.osrelease", str, &size, NULL, 0);
-      if (ret == 0) {
-         sscanf(str, "%d.%d.%d", &cMajor, &cMinor, &cInternal);
-      }
+      if (ret == 0) sscanf(str, "%hd.%hd.%hd", &version_[0], &version_[1], &version_[2]);
     }
-    *major = cMajor;
-    *minor = cMinor;
-    *internal = cInternal;
+    memcpy(k->version, version_, sizeof(version_));
 }
 
-bool IsKernelVersionUpperOrEqual(int major, int minor, int internal) {
-   int maj, min, inter;
-
-   GetKernelVersion(&maj, &min, &inter);
-
-   if ( maj > major) return true;
-   if ( maj == major) {
-      if ( min > minor ) return true;
-      else if ( min == minor && inter >= internal ) return true;
-   }
-   return false;
-}
-
-bool IsKernelVersionLessOrEqual(int major, int minor, int internal) {
-   int maj, min, inter;
-
-   GetKernelVersion(&maj, &min, &inter);
-
-   if ( maj < major) return true;
-   if ( maj == major) {
-      if ( min < minor ) return true;
-      else if ( min == minor && inter <= internal ) return true;
-   }
-   return false;
+int CompareKernelVersion(short int major, short int minor, short int component) {
+    struct kern k;
+    GetKernelVersion(&k);
+    if ( k.version[0] !=  major) return major - k.version[0];
+    if ( k.version[1] !=  minor) return minor - k.version[1];
+    if ( k.version[2] !=  component) return component - k.version[2];
+    return 0;
 }
 
 /*{
@@ -219,7 +200,7 @@ void ProcessList_goThroughEntries(ProcessList* super) {
        DarwinProcess_setFromLibprocPidinfo(proc, dpl);
 
        // Disabled for High Sierra due to bug in macOS High Sierra
-       bool isScanThreadSupported  = ! ( IsKernelVersionUpperOrEqual(17, 0, 0) && IsKernelVersionLessOrEqual(17, 4, 0));
+       bool isScanThreadSupported  = ! ( CompareKernelVersion(17, 0, 0) >= 0 && CompareKernelVersion(17, 5, 0) < 0);
 
        if (isScanThreadSupported){
            DarwinProcess_scanThreads(proc);


ylluminarious:
You sure you want to turn off this feature for all of High Sierra? [Some](https://github.com/hishamhm/htop/issues/682#issuecomment-361398071) [people](https://github.com/hishamhm/htop/issues/682#issuecomment-361925975) are reporting that XNU has been fixed by Apple in 10.13.4.

pmalhaire:
@ylluminarious not really, it seems that the issue is for versions 10.13.2 to 10.13.3 is it ?

anothernode:
@pmalhaire For me the issue started with macOS High Sierra 10.13.0 and persisted in all versions until 10.13.3. With the public beta of 10.13.4 that was released recently (and htop 2.0.2) I couldn't reproduce the issue anymore.

ilovezfs:
What is the value of `sysctl kern.osrelease` on 10.13.3 and on 10.13.4?

ylluminarious:
@ilovezfs On 10.13.3 it's `17.4.0`.

ilovezfs:
@pmalhaire so this PR also needs to exclude 17.4.0

ilovezfs:
@pmalhaire and >= 17.5.0 is safe again.

pmalhaire:
Thx I mixed up. Fixer @ilovefs @ylluminarious

pmalhaire:
Infortunately I don't have version to test the fix

ylluminarious:
@pmalhaire Thanks! Hopefully someone who is on the beta can try your patch and see if it works. Also @hishamhm should probably review this for merging.

darwin/DarwinProcessList.c
```diff
@@ -18,6 +18,55 @@ in the source distribution for its full text.
 #include <sys/mman.h>
 #include <utmpx.h>
 #include <err.h>
+#include <sys/sysctl.h>
+#include <stdbool.h>
+
+
+void GetKernelVersion(int ** major, int ** minor, int ** internal) {
```
mc10: Is the `**` needed here? I may be wrong, but it seems like you only need `int *major` etc. here if you change the bottom lines to be `*major = cMajor;` etc. and the variable declarations to be `int maj, min, inter;`.

pmalhaire: If I want to use pointers it's needed or the value won't be set.

Since the size of a pointer is the same or bigger than an int, I could use simple pointer.

pmalhaire: Fixing this.

pmalhaire: @mc10 should be fixed

anothernode:
I have a VM running High Sierra with one snapshot on _10.13.3_ and one snapshot on _10.13.4 public beta_. I'd be happy to run tests if it's of any help but I'd need some guidance on how and what exactly to test.

I could build htop from this branch (`pmalhaire:macos_keep_not_high_sierra_ok`) and test if it runs fine on both High Sierra versions but I wouldn't know how to test if that scan thread feature is actually switched on again on 10.13.4...

pmalhaire:
@anothernode
Actualy the function is giving thread infos si thread state or thread id are info which should be present when on 10.13.4 but not in 10.13.3 

https://github.com/hishamhm/htop/blob/8a147dd5b4bc35feebb14a3a55ee44186ce3cea4/darwin/DarwinProcess.c#L302

anothernode:
@pmalhaire All right, thanks!

So I compiled htop from `pmalhaire:macos_keep_not_high_sierra_ok` and let it run once on _10.13.3_ and once on _10.13.4 public beta_.

As you can see in the attached screenshots, on 10.13.3, the `S` column shows a `?` for the htop process but on 10.13.4 it shows an `R`. For all other processes though it shows `?` on either version. I hope that helps.

_High Sierra 10.13.3:_
screenshot-htop-high-sierra-10 13 3

_High Sierra 10.13.4 public beta:_
screenshot-htop-high-sierra-10 13 4-beta

pmalhaire:
Seems fine @anothernode

hishamhm:
Hey @pmalhaire! Thank you for this patch! I'm wondering though if it's worth keeping a lot of version detection code around to catch one bug in one specific point release... is 10.13.4 about to be released? I think once that is out, basically all High Sierra users will upgrade to it, right? So at that point, reenabling the feature unconditionally would make sense. What do you think? (I don't have a big understanding of typical macOS user upgrade patterns)

pmalhaire:
@hishamhm You are right most people will upgrade to a newer version. I didn't find updated upgrade rate but I d'say from past experiences that 30% of users will not upgrade 6 month after the upgrade is released.

Blocking it at package manager level (brew) is an other option what do you think @ilovezfs ?
something like 
``` 
if version is 10.13.3 
    error "please upgrade you sierra version to 10.13.4 or upper"   
```

Unfortunately macOs bug rate is quite high at the time. It may change, but you may need this kind of code again.

I'll compact my code anyway, it's too long for what it does.

ylluminarious:
@hishamhm @pmalhaire I know a fair number of people who do not upgrade right when a new release comes out. Even if they're already on the most recent major version. I am not one of those people most of the time, but I don't think that having some extra lines will hurt anything, considering the fact that people often have a good reason not to upgrade to a new minor version right away. Downtime, compatibility, bugs in the new version are all valid concerns and I think it's a bad practice to impose such minor versions on users, when it's not even much of a pain for the developer himself.

pmalhaire:
Thanks @ylluminarious for your explaination.

@hishamhm Given what @ylluminarious added I'd suggest merging this code.

ylluminarious:
@hishamhm Any reason this hasn't been merged yet? 10.13.4 beta 4 came out 2 days ago, which probably means that a release is imminent. I would merge this soon so that everything is back to normal for users of 10.13.4.

ylluminarious:
@hishamhm Surprisingly another beta came out a few days ago... there may be one more beta before the next release, but probably not. Please merge these changes soon or at least provide some feedback if something else needs done.

ylluminarious:
@mc10 Since you're the one assigned to review this PR, can you see anything else that needs to be done?

anothernode:
I wouldn't bet on the 5th beta being the last one. Both macOS [10.13.2](https://www.macrumors.com/2017/12/01/macos-high-sierra-10-13-2-root-fix/) and [10.13.3](http://osxdaily.com/2018/01/19/macos-10-13-3-beta-6-and-ios-11-2-5-beta-7-released-for-testing/) had seen 6 betas before the release and iOS 10.2.5 even had 7.

hishamhm:
Merged! Thank you @pmalhaire and sorry for the long delay in merging this! (I really hoped this would be a quick hiccup in High Sierra that would soon go away and I could just reenable the feature in the near future, but it doesn't seem so!...)

ylluminarious:
@hishamhm Thank you very much for merging this! I'm sure we can all appreciate the solution given by @pmalhaire.

ylluminarious:
@ilovezfs The homebrew formula for `htop` should probably be updated now.

ilovezfs:
@ylluminarious voilà https://github.com/Homebrew/homebrew-core/pull/25751

ylluminarious:
@ilovezfs @pmalhaire @hishamhm @anothernode Yikes, sorry folks, looks like this patch did not fix the issue. According to [my own testing,](https://github.com/hishamhm/htop/issues/682#issuecomment-377053296) this bug is still alive and well on High Sierra versions lower than 10.13.4.

ilovezfs:
I've reverted this patch from 2.1.0 in Homebrew: https://github.com/Homebrew/homebrew-core/pull/25887 until we figure out what went wrong here.

@pmalhaire I think the comparisons are backwards here. Won't `CompareKernelVersion(17, 5, 0)` be a positive number not a negative number on an affected system?

ylluminarious:
@ilovezfs @pmalhaire Sounds like a fine situation for good ol' `printf` and friends to tell us whether or not the comparison is actually working.

ilovezfs:
@ylluminarious are you suggesting we not guess?

ylluminarious:
@ilovezfs Heheh, well... yes :)

I've no problem with testing the reverse comparison patch that you made, but it sure wouldn't hurt to test it both ways and write some sort of logging statement indicating whether the main boolean expression is doing what we expect it to. Just as a sanity check, if nothing else.

Smattr:
Er, yes, the comparisons are is back-to-front, aren't they? Copy and pasting into an isolated program:

```c
#include <stdbool.h>                                                                                                                          [0/76]
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include <sys/sysctl.h>

struct kern {
    short int version[3];
};

void GetKernelVersion(struct kern *k) {
   static short int version_[3] = {0};
   if (!version_[0]) {
      // just in case it fails someday
      version_[0] = version_[1] = version_[2] = -1;
      char str[256] = {0};
      size_t size = sizeof(str);
      int ret = sysctlbyname("kern.osrelease", str, &size, NULL, 0);
      if (ret == 0) sscanf(str, "%hd.%hd.%hd", &version_[0], &version_[1], &version_[2]);
    }
    memcpy(k->version, version_, sizeof(version_));
}

int CompareKernelVersion(short int major, short int minor, short int component) {
    struct kern k;
    GetKernelVersion(&k);
    if ( k.version[0] !=  major) return major - k.version[0];
    if ( k.version[1] !=  minor) return minor - k.version[1];
    if ( k.version[2] !=  component) return component - k.version[2];
    return 0;
}

bool foo() {
    bool isScanThreadSupported  = ! ( CompareKernelVersion(17, 0, 0) >= 0 && CompareKernelVersion(17, 5, 0) < 0);
    return isScanThreadSupported;
}


int main(void) {
    puts(foo() ? "true" : "false");
    return EXIT_SUCCESS;
}
```

This tells me "true" and I'm on macOS 10.13.3 so it should be telling me "false." Should this be the following?

```c
bool isScanThreadSupported  = ! ( CompareKernelVersion(17, 0, 0) < 0 && CompareKernelVersion(17, 5, 0) >= 0);

**Comment by ilovezfs** [comment]
No, I think it needs to be
```
    bool isScanThreadSupported  = ! ( CompareKernelVersion(17, 0, 0) <= 0 && CompareKernelVersion(17, 5, 0) > 0);
```
otherwise 10.13.4 will be disqualified too.

Or we can reverse the subtraction order in CompareKernelVersion.

ylluminarious:
@ilovezfs @Smattr I know this still sounds elementary and obvious... but this is a place where I would use `printf` or similar to verify that the conditional is working correctly. We definitely need some assurance as to which conditional is correct since my head is starting to spin after thinking about this logic several times... :P

After [some unsuccessful testing](https://github.com/hishamhm/htop/issues/682#issuecomment-377133691) in a VM, I'd like to have more confidence before being a guinea pig again. I am sure that others would like the same assurance. I would do it myself, but I have not had time to go and compile `htop` and apply these patches manually.

ilovezfs:
@ylluminarious I did, which is how I identified the off-by-one error in @Smattr's suggestion. But please feel free to verify my findings yourself :)

ylluminarious:
@ilovezfs Oh cool, that's great. I didn't realize you had already done that. Thank you, I have some more confidence to try it out now. I suppose I can try it out at some point tomorrow or the next day unless someone beats me to it :)

ilovezfs:
@ylluminarious please note that my suggestion above is different from what I said in the other issue, which is what accounts for the failure monouser7dig still saw.

ylluminarious:
@ilovezfs Right, I noticed that it was different. Thank you for further clarifying that.

pmalhaire:
I'll have a look

pmalhaire:
I definetly switched the test while compacting my code, I'll fix and send a test executable tonight.

anothernode:
Personally I'm happy with always running `htop` with `sudo` until macOS 10.13.4 gets released.

So maybe we could also consider a kind of patch where `htop` just refuses to start and print some explanatory error message instead, if invoked without `sudo` on macOS `10.13.0 >= x <= 10.13.3`?

It kind of seems to be a bit of a waste of effort to put much more time and energy into this, now that we know that 10.13.4 will fix the actual root cause of the whole mess...

ilovezfs:
@anothernode It's cool that you have root access on your computer. Many users don't.

pmalhaire:
[test.c.txt](https://github.com/hishamhm/htop/files/1862022/test.c.txt)
It should be fine now. ^ This is the test with all versions ^

Sorry for the error.

ilovezfs:
Apple has now released 10.13.4

ylluminarious:
Ok, so I attempted to test @ilovezfs's fix yesterday, but I had some issues connecting to https://hisham.hm, so I could not get the modified formula to install. Moreover, I noticed that @pmalhaire started a new PR at #772, so which patch should be tested now?

pmalhaire:
@ylluminarious @ilovezfs the patch I made just fixes the error I made. It's the same as what @ilovezfs suggested.

ylluminarious:
@pmalhaire Ok, your solution is slightly different in that it changes the arithmetic in `CompareKernelVersion` rather than changing the comparison operators. But I see that it is *effectively* the same by testing out both versions from you and @ilovezfs. I guess I'll try out your patch in #772 and see what happens.

ylluminarious:
@pmalhaire @ilovezfs @hishamhm Looks like #772 does the job and fixes this problem for good.

PR #729, opened 2018-02-05
Add <sys/sysmacros.h> to two files.

Polynomial-C:
Future glibc releases will no longer include sysmacros implicitly.

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -19,6 +19,9 @@ in the source distribution for its full text.
 #include <sys/param.h>
 #include <sys/stat.h>
 #include <sys/types.h>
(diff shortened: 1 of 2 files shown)


hishamhm:
Hi! Thank you for the PR.

Travis CI says this builds to fail on the Mac: `<sys/sysmacros.h>` is not available on macOS so it shouldn't be in `Process.c`. I'd prefer to avoid OS-specific ifdefs in the root directory files and leave OS-specific code to their subdirectories. Any suggestions?

Explorer09:
@Polynomial-C 
Would you please try #746 in which I fixed it a better way?

Polynomial-C:
Closed in favor of #746

PR #731, opened 2018-02-07
linux/LinuxProcessList: fix reading of number of read syscalls of pro…

marckleinebudde:
…cess

The "if" tests if the character at index "5" is 'r', as a first quick
check. However at index "5" will always be a colon ":". This patch fixes
the off-by-one error. htop now shows proper values in the RD_SYSC
column.

Signed-off-by: Marc Kleine-Budde <mkl@pengutronix.de>


hishamhm:
Thank you for the fix!

PR #741, opened 2018-02-18
Import Solaris support

gmbroome:


hishamhm:
@gmbroome Thanks for sharing your branch! The other PR, #740, contains the patch you sent me by email, in which I made some formatting fixes. If this PR contains the most up-to-date code, could you apply the same indentation fixes of that PR to your branch as well? Thanks!

gmbroome:
@hishamhm Indentation fixes applied as requested - thanks for being polite about my indiscriminate mixing of tabs and spaces.

hishamhm:
Thanks! Please also remove the SolarisProcessList_scan stub function (I'll remove it from `unsupported/` as well.

gmbroome:
@hishamhm SolarisProcessList_scan() stubs removed from solaris/SolarisProcessList.c and solaris/SolarisProcessList.h

hishamhm:
@gmbroome I asked around for some beta-testing on Twitter, and got this response: https://twitter.com/HummyPkg/status/965358173728268289
It seems to have been based on #740 though, so it might have been fixed already in this branch.

gmbroome:
@hishamhm Correct, the referenced crash has already been corrected in PR#741.  I was experiencing that myself at one point.

gmbroome:
@hishamhm Also worth noting re: the twitter post - Illumos and derivatives are my main concern.  I'm not even testing on Oracle Solaris proper.  I'd even be 100% okay with renaming the platform in htop from Solaris to Illumos if that would make more sense.

hishamhm:
> I'd even be 100% okay with renaming the platform in htop from Solaris to Illumos if that would make more sense.

Now that's a very interesting question! I wouldn't mind that. (Go free software!)

hishamhm:
@gmbroome so, what do you say? `solaris/` or `illumos/`? Your call. And then I think we're good to merge!

gmbroome:
@hishamhm I'm on the fence, but let's leave it Solaris for now.  At the moment, `uname -a` on all Solaris 11 and Illumos platforms still reports "solaris2.11".  Until / unless Illumos decides to no longer call itself Solaris internally, trying to detect and support Illumos discretely is probably just bikeshedding.

hishamhm:
Merged, finally! Sorry about the long process, and thank you!!

gmbroome:
@hishamhm Yay!  Consider the champagne uncorked.  Now the real work begins- improving and supporting the port.  Thank you very much for all of the help and guidance to date.

papertigers:
@ninefathom hey thanks for your work supporting htop proper without lx proc.
One thing I would like to see changed is how you decided to calculate the Mem graph. You now report system memory rather than a zones RSS cap which is rather unfortunate in my opinion. It does look like swap is calculated correctly though.  I haven't looked at where you grab that info from but all of it is in kstats.

ninefathom:
@papertigers Memory reporting, in general, is on my "needs to be improved" list.  In addition to the system vs. zone-specific observation you made, a rational mapping of Solaris memory usage types to the various types in htop (based off of Linux originally) is not obvious to me.  I made a "best guess" to get the port working, but it's quite ugly.

Anyway, your zone-specific memory reporting request seems quite reasonable, is something I could use myself, and (I suspect) would be straightforward enough.  I'll prioritize that in terms of "stuff to fix in the port".

P.S. the next biggest thing vexing me is that CPU time per process is not correct... or more precisely, it doesn't match the output of prstat.  The time reported for a process (i.e. a representative LWP) is not cumulative for all of the LWPs in the process, and I can't seem to find a straightforward way to make it so.  The way it's done in libproc for e.g. prstat won't work for htop due to differences in how the process lists are treated in each, and every approach I've tried ends up with ugly, problem-prone code.

despair86:
from what i've been able to observe between the various Solaris forks is that they are *largely* binary compatible, if you compile with GNU CC across any of them, there *shouldn't* be any major problems.

Sun Studio 12.6 requires patches only to `SUNWms[ru]` (the advanced maths library) and `libumem` (can't recall what package provides this atm), to run on any of the libre distributions. You would also need to link the C++ABI library (here, called `libCrun`) *statically*

cheers
-rick

PR #749, opened 2018-02-26
Fix bashisms

Jesin:
The configure script relied on bash-specific extensions to shell syntax
and behavior, causing build failures on systems with other /bin/sh
implementations. This commit replaces those with equivalent constructs
that should work in all POSIX shells.


hishamhm:
@Jesin Thank you! 

A curiosity I've always had: in which contemporary systems is adding `x` to both sides of a variable test still necessary? In every non-bash shell I tested, `"$empty_var"` was properly expanded to a zero-length token.

Jesin:
@hishamhm Zero-length tokens are usually fine. But what happens when the token turns out to be `-n`, or `!`, or `=`, or `--help`, or anything else syntactically meaningful? This is what varies between implementations.

None of the many implementations of `test` have ever read syntactic meaning into any token that begins with `x`, so prepending `x` avoids this entire class of problems.

PR #750, opened 2018-02-28
Changed type in StringUtils.c to avoid integer overflows

sdicke:
Some return values from functions that return size_t were assigned to variables from type int. There are cases where sizeof(size_t) > sizeof(int). There are malloc calls influenced by these variables.

### DIFF
--- a/StringUtils.c
+++ b/StringUtils.c
@@ -27,11 +27,11 @@ in the source distribution for its full text.
  */
 
 char* String_cat(const char* s1, const char* s2) {
-   int l1 = strlen(s1);
-   int l2 = strlen(s2);
+   size_t l1 = strlen(s1);
+   size_t l2 = strlen(s2);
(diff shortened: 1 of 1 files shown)


hishamhm:
This does look correct, but it seems it would only happen in pretty pathological cases (like strings over 2GB in length), right?

wataash:
I agree this changes because:

- I don't consider this overflow as unrealistic. There's no reason to believe htop will never be run on an architecture whose int has a size of 16 bits.
- Bugs might be generated when a piece of this code is copy-pasted.
- [CERT recommends it.](https://wiki.sei.cmu.edu/confluence/display/c/INT01-C.+Use+rsize_t+or+size_t+for+all+integer+values+representing+the+size+of+an+object)

hishamhm:
Fair enough, merging it! :)

PR #751, opened 2018-03-01
MakeHeader.py: ignore unicode characters since .h files are acsii-enc…

Polynomial-C:
…oded.

Header creation fails with non-utf8 locale and python3.
Simply set LC_ALL="C" and use python3 to reproduce the issue.

### DIFF
--- a/scripts/MakeHeader.py
+++ b/scripts/MakeHeader.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-import os, sys, string
+import os, sys, string, io
 try:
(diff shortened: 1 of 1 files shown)


Polynomial-C:
See downstream bug report: https://bugs.gentoo.org/646880

Explorer09:
Wait a minute. IIRC there _are_ UTF-8 characters in several .c files in htop. Why did you say to ignore them?

hishamhm:
This was superseded by #770. Thanks for bringing up the issue!

PR #753, opened 2018-03-05
Solaris: assorted fixes

gmbroome:
Added proper backtrace on abort, plus linking against libmalloc - a simple thing that I should have seen earlier, which solves quite a few crashes.  Small pull, major improvement.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -243,6 +243,7 @@ fi
 
 if test "$my_htop_platform" = "solaris"; then
    AC_CHECK_LIB([kstat], [kstat_open], [], [missing_libraries="$missing_libraries libkstat"])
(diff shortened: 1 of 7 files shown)


gmbroome:
@hishamhm Courtesy of Peter Tribble bringing ncurses6 into Tribblix over the weekend, it became much easier for me to do side-by-side 64-bit/32-bit today, which it turns out is really handy for exposing memory allocation issues.  I was doing all sorts of superbly unintelligent memory allocation things, particularly around string handling.  As of commit afd370b, stability seems much better.

hishamhm:
Merged, thank you! Looks like we're good for making a release!

PR #757, opened 2018-03-07
Use fork/exec instead of popen to run lsof

### LINKED ISSUE #675: lsof exit code not detected correctly
I've tried using `l` and `s` on a box with neither `lsof` nor `strace` installed and noticed that while I got an error message for `s`, there was none for `l`.  This seems to be the fault of how both programs are spawned, for `lsof` a pipe is opened with `popen(3)` and a return value of null is interpreted as exit code 127.  This is wrong as `popen(3)` starts a shell and failure other than memory allocation is related to whether the shell could be opened.  Furthermore the return value of `pclose(3)` is interpreted as the `lsof` exit code which is wrong as well, according to `wait(3)` it contains far more information than that to allow telling whether the process did exit normally or by a signal, you'll have to use the `WIFEXITED` and `WEXITSTATUS` macros to extract it from that value.  I've created a small example demonstrating this in a hopefully portable way:

```c
#include <stdio.h>
#include <sys/wait.h>

int main() {
   FILE* fd = popen("not-lsof", "r");
   int ret, exit_code;

   if (fd) {
      printf("Success!\n");
      ret = pclose(fd);
      if (ret < 0) {
         printf("Failed closing fd\n");
      } else {
         if (WIFEXITED(ret)) {
            printf("Process did exit with %d\n", WEXITSTATUS(ret));
         } else {
            printf("Process terminated by a signal\n");
         }
      }
   } else {
      printf("Fail!\n");
   }

   return 0;
}
```

I'm not sure how to rework `OpenFilesScreen.c` so that it takes the more precise error handling into account and would be happy about any suggestions.

wangqr:
`popen` will run the command with `sh -c`, which makes it hard to detect errors and get the return value. This PR rewrites the code to run `lsof` using `fork`/`exec` instead of `popen` to get more detailed result.

This PR fix #675.

### DIFF
--- a/OpenFilesScreen.c
+++ b/OpenFilesScreen.c
@@ -76,16 +76,35 @@ void OpenFilesScreen_draw(InfoScreen* this) {
 }
 
 static OpenFiles_ProcessData* OpenFilesScreen_getProcessData(pid_t pid) {
-   char command[1025];
-   xSnprintf(command, 1024, "lsof -P -p %d -F 2> /dev/null", pid);
-   FILE* fd = popen(command, "r");
+   char buffer[1025];
+   xSnprintf(buffer, 1024, "%d", pid);
    OpenFiles_ProcessData* pdata = xCalloc(1, sizeof(OpenFiles_ProcessData));
    OpenFiles_FileData* fdata = NULL;
    OpenFiles_Data* item = &(pdata->data);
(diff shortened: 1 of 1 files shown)


hishamhm:
Merged, thank you!

PR #758, opened 2018-03-07
Include pkg.m4 into htop source tree.

Explorer09:
Without pkg.m4, and when the developer's system did not have pkg-config
properly configured, `autoreconf` can silently generate a broken
configure script without failing. (The configure script will produce a
syntax error on `PKG_PROG_PKG_CONFIG()` when run.)

The fix to this is including pkg.m4 into htop source. This M4 file can
be used and updated independently of user system's pkg-config version.

The pkg.m4 version included here is 0.29.2.


hishamhm:
I'm unsure about this one... Don't we risk to break the build in other people's systems which _do_ have pkg-config installed and happens to be a different version, incompatible with the bundled pkg.m4?

Explorer09:
@hishamhm 
The break would only happen if the user have pkg-config that's _newer_ than the bundled pkg.m4, and the pkg.m4 writers decide to break the compatibility in the future. (But only M4 macros' compatibility; the macro snippets are otherwise independent of user's installed pkg-config version.)

I don't think the breakage argument holds because AFAIK the pkg.m4 would be updated only when the user rebuilds `configure` through `autogen.sh`. pkg.m4 has no direct business with users building from a released (or pre-packaged) tarball.

So pkg.m4 version matters more with us packagers than with the builders.

I see one pro of including pkg.m4 in the source tree: we would never ship with older pkg-config checking code (in `configure`) accidentally (when we packagers' build machine contains an older pkg-config).
And the con is: we need to periodically check and update pkg.m4 to keep in sync with the checking code changes upstream.

Explorer09:
P.S. I personally would like to avoid the stupid "ifdef PKG_PROG_PKG_CONFIG" check like this code, when I know pkg.m4 macros can expand independently of the packager's installed pkg-config version. 
https://github.com/traviscross/mtr/blob/e7f17cb54d6afba663a982d60f169027b02c7dd9/configure.ac
```
m4_ifndef([PKG_PROG_PKG_CONFIG],
  [m4_fatal(
[Could not locate the pkg-config autoconf macros.  These are usually located
in /usr/share/aclocal/pkg.m4.  If your macros are in a different location,
try setting the environment variable ACLOCAL_OPTS="-I/other/macro/dir"
before running ./bootstrap.sh again.])
])
```

hishamhm:
Lots of people build from Git master, so I don't see `autogen.sh` as being only for distro package maintainers.

I would rather do the more conservative PKG_PROG_PKG_CONFIG check instead.

PR #760, opened 2018-03-07
Use AM_CFLAGS and AM_LDFLAGS in Makefile.am

Explorer09:
This reduces generated Makefile.in size by 74%.
(217319 bytes -> 56326 bytes)

Automake considers that `<prog>_CFLAGS` and `<prog>_LDFLAGS` are
program-specific build rules, and when such are specified, Automake
will generate additional code just to avoid the "generic" and
package-wide AM_CFLAGS or AM_LDFLAGS. (Especially for `<prog>_CFLAGS`,
Automake will rename generated object files to become "prog-foo.o" and
such, and it's _a lot_ of code to achieve this in Makefile.)

There's no reason for htop to rename intermediate object files. It's
better to make things simpler.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,8 +12,8 @@ applications_DATA = htop.desktop
 pixmapdir = $(datadir)/pixmaps
 pixmap_DATA = htop.png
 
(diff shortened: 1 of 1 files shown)


hishamhm:
Nice, thank you!

I honestly don't recall if I did that for some particular reason or if I just followed some Autotools tutorial which did that 14 years ago (most likely the latter) :D

PR #763, opened 2018-03-17
New makefile targets to rebuild and clean htop headers.

Explorer09:
`make htop-headers` will regenerate all '.h' headers in htop source for
all platforms.
`make clean-htop-headers` will delete all generated htop headers.

Because of the introduction of these two targets, I slightly changed
the style of platform-specific portions of makefile rules.
Please comment if you accept such a style, or need me to revert to old
makefile style.


hishamhm:
That style works, thank you!

PR #765, opened 2018-03-22
Solaris: Fix virt and resident memory sizes

gmbroome:
Solaris platform code was incorrectly parsing the resident and virtual memory sizes from Solaris procfs.  The original system for the Solaris port was a SPARC T2+ with 8 KiB pages, but Solaris on x86 generally uses 4KiB pages.  The code has been revised to detect page size at run time, and report per-process usage correctly to the platform-independent code.


gmbroome:
Had a last-minute "wait, is that right" but it is, in fact, correct.  Output of htop memory usage stats now aligns with that given by prstat.

PR #767, opened 2018-03-26
Don't generate 'configure' if PKG_PROG_PKG_CONFIG is undefined

Explorer09:
(Follow-up of pull request #758 )

Without pkg.m4, and when the developer's system did not have pkg-config
properly configured, `autoreconf` could silently generate a broken
configure script without failing. (The configure script would produce a
syntax error on `PKG_PROG_PKG_CONFIG()` when run.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4,6 +4,9 @@
 AC_PREREQ(2.65)
 AC_INIT([htop],[2.1.0],[hisham@gobolinux.org])
 
+m4_ifndef([PKG_PROG_PKG_CONFIG],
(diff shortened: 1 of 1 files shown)


hishamhm:
Since we are only using pkgconfig with `--enable-delayacct`, would it be possible to check that conditionally? (i.e. would that work inside the `if test "x$enable_delayacct" = xyes` block?)

Explorer09:
@hishamhm 
No, this check is done during the _generation_ of the configure script (that is, when running autogen.sh), not during running of the configure script. This added m4_ifndef macro adds no code to the configure script. It's just to make autogen.sh fail early if something goes wrong.
It's what you requested in #758, remember?

PR #768, opened 2018-03-27
Solaris: Implement LWP enumeration

gmbroome:
Since Solaris does not have system-wide unique identifiers per LWP, this change required creating a "virtual PID" for ProcessList, while maintaining the real PIDs and LWPIDs discretely per-process for display.  This change adds the new default LWPID column for Solaris.  Shadowing kill() in solaris/Platform.h was also needed to account for the virtual PIDs.

### DIFF
--- a/solaris/Platform.c
+++ b/solaris/Platform.c
@@ -146,8 +146,19 @@ void Platform_getLoadAverage(double* one, double* five, double* fifteen) {
 }
 
 int Platform_getMaxPid() {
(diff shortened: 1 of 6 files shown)


hishamhm:
Thanks! We should be able to get rid of the `kill()` trick once I devise a better abstraction in the cross-platform part. At first I was thinking of adding `Process_pid()`, `Process_ppid()` and `Process_tid()` macros, but the real issue may be that the cross-platform part assumes that pid/ppid can always be used as the tree-sorting keys.

PR #770, opened 2018-03-28
MakeHeader.py: Fix for non-utf8 environments

volkov-am:
Header creation fails with non-utf8 locale and python3.
Simply set LC_ALL="C" and use python3 to reproduce the issue.

env LC_ALL="C" ./scripts/MakeHeader.py MetersPanel.c
Traceback (most recent call last):
  File "./scripts/MakeHeader.py", line 32, in <module>
    for line in file.readlines():
  File "/usr/lib64/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 956: ordinal not in range(128)

This changes is python2 and python3 compatible

cStringIO.StringIO module is removed because it is not able to accept unicode strings
https://docs.python.org/2/library/stringio.html#cStringIO.StringIO

### DIFF
--- a/scripts/MakeHeader.py
+++ b/scripts/MakeHeader.py
@@ -1,12 +1,9 @@
 #!/usr/bin/env python
-import os, sys, string
+import os, sys, string, io
 try:
-   from cStringIO import StringIO
+   from StringIO import StringIO
 except ImportError:
-   try:
-      from StringIO import StringIO
-   except ImportError:
-      from io import StringIO
+   from io import StringIO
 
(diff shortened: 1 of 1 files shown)


volkov-am:
This pull request supersede #751

hishamhm:
@volkov-am Thank you for coming up with the solution! This is now merged. A release is coming this weekend!

PR #771, opened 2018-03-29
Solaris: misc. fixes. improvements, and cleanup

gmbroome:
Solaris port change summary:
- Removed all direct interaction with procfs files and dirs in favor of libproc usage (MAJOR CHANGE)
- Implemented process environment display
- Consolidated separate process vs LWP info-gathering loops into one (MAJOR CHANGE)
- Assorted cosmetic improvements

NOTE: this PR modifies configure.ac to add a new library dependency (libproc) if the Solaris platform is detected, alongside the existing libkstat and libmalloc dependencies.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -243,6 +243,7 @@ fi
 
 if test "$my_htop_platform" = "solaris"; then
    AC_CHECK_LIB([kstat], [kstat_open], [], [missing_libraries="$missing_libraries libkstat"])
+   AC_CHECK_LIB([proc], [Pgrab_error], [], [missing_libraries="$missing_libraries libproc"])
    AC_CHECK_LIB([malloc], [free], [], [missing_libraries="$missing_libraries libmalloc"])
 fi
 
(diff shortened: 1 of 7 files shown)


gmbroome:
@hishamhm With this PR, I think Solaris is about as feature-complete as it's going to get.

The last missing portable item is battery status.  I can't find any usable, standard, cross-architecture API for battery info on Solaris- prtdiag only works for OEM hardware (like the batteries on HBAs and such), acpidrv is both third-party and x86-only, etc.- and I can't imagine a huge need for this anyway.  So I'm going to have to call the lack of battery status an EWONTFIX item unless some more savvy Solaris admin knows of a clever solution.

lsof and strace replacements are available if those end up in the platform-specific code some day.

Unless something changes in the Solaris situation or portable features, from this point on I think any changes I do will be bug fixes or minor improvements.

gmbroome:
Yay!  Memory leak fixed.  Running instances on sparc and x86 showing stable memory usage for 2+ days.  Thanks to libumem, mdb, and ::findleaks.

hishamhm:
Merged manually after squashing a few bugfix commits, thank you! Release coming this weekend, no postponing from my part this time! :)

PR #772, opened 2018-03-29
osx : fix the inverted version test

pmalhaire:
switched flags for this https://github.com/hishamhm/htop/pull/728

### DIFF
--- a/darwin/DarwinProcessList.c
+++ b/darwin/DarwinProcessList.c
@@ -38,12 +38,17 @@ void GetKernelVersion(struct kern *k) {
     memcpy(k->version, version_, sizeof(version_));
 }
 
+/* compare the given os version with the one installed returns:
+0 if equals the installed version
+positive value if less than the installed version
+negative value if more than the installed version
+*/
 int CompareKernelVersion(short int major, short int minor, short int component) {
     struct kern k;
     GetKernelVersion(&k);
-    if ( k.version[0] !=  major) return major - k.version[0];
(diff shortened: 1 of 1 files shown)


Explorer09:
Hello. I wonder what's the reason for inverting the function behavior of CompareKernelVersion() ?
Also, when we are improving the function, is it okay to add `static` qualifier to CompareKernelVersion? If you are familar with C you know what it means.

pmalhaire:
@Explorer09 I do invert the function because I made a mistake while writing it. See https://github.com/hishamhm/htop/pull/728
It makes sense tout make it static.

pmalhaire:
@hishamhm I accidentaly inverted the test when I compactes the code. This is a fix for this

ylluminarious:
@pmalhaire @Explorer09 After testing this patch on my own and verifying the test with some logging, I can confirm that this patch fixes the problem and it should no longer occur. Unless, of course, something other than `DarwinProcess_scanThreads` is causing the problem, but there is no evidence to indicate that.

ylluminarious:
I applied this patch and left the resultant program running for **4 hours** on 10.13.3 and didn't crash... I suggest that this patch be merged as soon as possible.

hishamhm:
Merged! Thanks, @pmalhaire! A new release is coming this weekend!

ylluminarious:
@hishamhm Thanks again! Sorry for the false relief in the previous PR... but thanks for your patience!

PR #775, opened 2018-04-06
Require pkg-config as a build dependency on Linux only

### LINKED ISSUE #774: ./configure Error on OpenIndiana
My system: OpenIndiana The Illumos Project SunOS 5.11 illumos-17f11284b4 April 2018

Just run autogen.sh - perfect

After that ./configures throws 

> ./configure: line 6745: syntax error at line 6751: `(' unexpected

needed to uncomment the affected if clause

```
if test "x$enable_delayacct" = xyes
then
   PKG_PROG_PKG_CONFIG()
   PKG_CHECK_MODULES(LIBNL3, libnl-3.0, , missing_libraries="$missing_libraries libnl-3")
   PKG_CHECK_MODULES(LIBNL3GENL, libnl-genl-3.0, , missing_libraries="$missing_libraries libnl-genl-3")
   CFLAGS="$CFLAGS $LIBNL3_CFLAGS $LIBNL3GENL_CFLAGS"
   LIBS="$LIBS $LIBNL3_LIBS $LIBNL3GENL_LIBS"

$as_echo "#define HAVE_DELAYACCT 1" >>confdefs.h

fi
```

After that `./configure` as well as `make `and `make install` worked.
Binary also works.

hishamhm:
This is a follow-up of #767.

@Explorer09 this is what I meant by enabling it conditionally. I believe that doing it this way, we make pkgconfig a build dependency on Linux only.

This should fix #774 without adding a new dependency. @schenklklopfer, could you test this branch? Thank you!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -281,15 +281,19 @@ then
    AC_DEFINE(HAVE_SETUID_ENABLED, 1, [Define if setuid support should be enabled.])
 fi
 
-AC_ARG_ENABLE(delayacct, [AS_HELP_STRING([--enable-delayacct], [enable linux delay accounting])],, enable_delayacct="no")
+AC_ARG_ENABLE(delayacct, [AS_HELP_STRING([--enable-delayacct], [enable Linux delay accounting])],, enable_delayacct="no")
 if test "x$enable_delayacct" = xyes
 then
-   PKG_PROG_PKG_CONFIG()
-   PKG_CHECK_MODULES(LIBNL3, libnl-3.0, [], [missing_libraries="$missing_libraries libnl-3"])
-   PKG_CHECK_MODULES(LIBNL3GENL, libnl-genl-3.0, [], [missing_libraries="$missing_libraries libnl-genl-3"])
-   CFLAGS="$CFLAGS $LIBNL3_CFLAGS $LIBNL3GENL_CFLAGS"
-   LIBS="$LIBS $LIBNL3_LIBS $LIBNL3GENL_LIBS"
-   AC_DEFINE(HAVE_DELAYACCT, 1, [Define if delay accounting support should be enabled.])
+   m4_ifdef([PKG_PROG_PKG_CONFIG], [
+      PKG_PROG_PKG_CONFIG()
+      PKG_CHECK_MODULES(LIBNL3, libnl-3.0, [], [missing_libraries="$missing_libraries libnl-3"])
+      PKG_CHECK_MODULES(LIBNL3GENL, libnl-genl-3.0, [], [missing_libraries="$missing_libraries libnl-genl-3"])
+      CFLAGS="$CFLAGS $LIBNL3_CFLAGS $LIBNL3GENL_CFLAGS"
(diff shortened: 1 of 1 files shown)


Explorer09:
>
> @Explorer09 <https://github.com/Explorer09> this is what I meant by
> enabling it conditionally. I believe that doing it this way, we make
> pkgconfig a build dependency on Linux only.
>
Sorry, Hisham, but that won't work as you expected. The reason is that the
shell's `if` conditional is evaluated only when _running_ the configure
script, not the _generation_ of it, so the net result will be the same as
my fix -- you'll always depend on pkg.m4 when generating the configure.

There have been only two approaches to the problem:
1. Supply pkg.m4 by ourselves when user checks out from git repo. (If
building from release tarball, the pkg.m4 macro will be supplied already in
the aclocal process, so the problem doesn't apply.)
2. Require the user to have a copy of pkg.m4 installed when checking out
from git.

Since the first is rejected, I propose the second in the PR, there is no
other way.
(Not unless you tell user to remove the pkg-config macros from configure.ac
and make them aware of what they are doing.)

>

Explorer09:
PS. I have been recommending the first approach since it would be easier and frees the users from requiring pkg-config during the bootstrap phase (the "bootstrap" of the configure script).
Since the configure script has to be deterministic (especially for a htop release, i.e. for `make dist`), I can't make pkg.m4 an optional dependency. When we use it, we have to make one of the two choices I stated before: Either we supply it, or require user to do so.

hishamhm:
@Explorer09 Ok, I just tested this without pkg-config installed in the system and it does work as I intended. (I assume you just read the diff and didn't test this.)

What this PR does is:

* if pkgconfig is installed, it will expand the code inside the shell `if`, adding the pkgconfig-based tests for the dependencies of Linux delay accounting.
* if pkgconfig is not installed, it will add an error message inside the test of Linux delay accounting, telling the user to install pkgconfig and rebuild the configure script if they want to use Linux delay accounting.

The end result is:

* people running Linux
   * will not need pkgconfig when not using delay accounting
   * will need pkgconfig when using delay accounting
      * if they don't have it
        * they are told by configure they need to install it and re-run autogen.sh when running from Git
        * they are told by configure they need to install it and re-run configure when running from the tarball
* people not running Linux
   * will never need pkgconfig

...and in none of the above scenarios the generated `configure` script produces unexpanded macros.

I'm happy with this!

> there is no other way. 

https://www.youtube.com/watch?v=dMjQ3hA9mEA

Explorer09:
@hishamhm
Well, I guess I mis-read the commit diff a bit. But even I reread it correctly, your fix this way would bring the problem into `make dist`.

hishamhm:
`make dist` is essentially for my own use for producing the release tarballs. It is fair to assume that pkgconfig is a hard dependency in that scenario. I'm concerned about users running straight from a Git checkout.

Explorer09:
I thought it would be better to have something to check pkg-config in `make dist` after this change.
Maybe I should propose a dist-hook after this PR.
(I'm sort of a perfectionist here.)

PR #776, opened 2018-04-07
Disable 'make dist' when pkg.m4 is unused in configure

Explorer09:
Follow-up of #775.
This would prevent a careless future package maintainer (forks or distros) from creating a release tarball with a defective configure script. :)

Also, add a warning in the autogen.sh phase if pkg.m4 is unused.


hishamhm:
Great, thanks! (Sorry, this missed 2.2.0 by a tiny bit.) There seems to be a spurious character in the patch: (showed up as 　 -- I'm getting a Unicode box in this machine)

hishamhm:
Merged, thank you!

PR #773, opened 2018-04-02
Add tree view command line flag

lytedev:
Hello!

I know there wasn't any talk of this beforehand, but I wanted to propose a `-t --tree` flag to fire up htop in the tree view by default. This is my naive initial implementation and I'm more than open to guidance going forward. I can see there might be a conflict with the `-s` flag, so I'm not sure how we'd reconcile that.

If there was any previous discussion regarding something similar, I do apologize. I admit I only spent a few minutes perusing the existing (closed) feature request issues.

Thanks for your time!

PR #777, opened 2018-04-07
Add tree view flag to man page

lytedev:
This is on the tails of https://github.com/hishamhm/htop/pull/773, adding the new tree view command line flag to the man page.


hishamhm:
Awesome, thanks!

PR #780, opened 2018-04-10
fix option string

eworm-de:
This broke with commit db05ba61065f64b59d0014518be0786b5439e54c.


hishamhm:
Thank you for the fix!

PR #781, opened 2018-04-12
Solaris: Make last change to Platform.h persistent

ninefathom:
Fix a "whoops" - last change to solaris/Platform.h gets clobbered when that file is regenerated.  This makes the change persistent.


gmbroome:
Also addresses part of issue #783 by cleanly failing during configure on Solaris 10 with explanation

hishamhm:
Merged, thanks!

PR #789, opened 2018-04-27
Fix #define to match header when regenerating with MakeHeader.py

fasterit:
from Debian https://sources.debian.org/src/htop/2.2.0-1/debian/patches/fix-linux-process.patch/

The LinuxProcess.h and LinuxProcess.c don't match which means htop 2.2.0 only compiles until you regen the header files. This fixes the inconsistency from 47cf1532b0c9fbc70bada5022a7db07d3cc4811a.
/DLange

### DIFF
--- a/linux/LinuxProcess.c
+++ b/linux/LinuxProcess.c
@@ -143,7 +143,7 @@ typedef struct LinuxProcess_ {
 } LinuxProcess;
 
 #ifndef Process_isKernelThread
-#define Process_isKernelThread(_process) ((LinuxProcess*)(_process)->isKernelThread)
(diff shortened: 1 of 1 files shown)


Explorer09:
@fasterit The headers will be all re-generated by MakeHeader.py in the future, so all inconsistencies will be eliminated during the process. What I'm saying is: **Your commit message about "to match header" will be unclear.** It's better that you state about **the missing parentheses** and that's a regression of 47cf1532b0c9fbc70bada5022a7db07d3cc4811a instead.

hishamhm:
@fasterit merged, thank you!

PR #790, opened 2018-04-29
Fix parsing -s on the command line after db05ba610

haasn:
The named commit added an option parameter to -t, but seems to have
mistakenly added the `:` in the wrong place, thus breaking the parsing
of `-s` (segfaults with optarg=NULL).


haasn:
Actually, looking at the commit in question, it doesn't make sense that there would be a `:` at all, since `-t` does not accept a parameter. I've force-pushed a new version that removes the `:`.

hishamhm:
Merged the other PR, thank you!

PR #811, opened 2018-07-07
htop.c: remove unused "--io" / "-i"

LionNatsu:
Introduced from https://github.com/hishamhm/htop/commit/3383d8e5561dfc6fb2b65e0a194df94ccb5e08af (2.0.0) but never used.


hishamhm:
Nice catch — a more general feature in that direction is coming in the `next` branch for 3.0!

PR #812, opened 2018-07-10
Fix process name updates for shorter strings

ScoreUnder:
When a process name changes from a long string to a short string,
truncate instead of just overwriting the beginning.

# Reproduction steps

1. Run htop with a reasonable update frequency and `Update process names on every refresh` turned on.
2. Run a command which switches to a shorter name, e.g. `sh -c "sleep 5; exec vim"`
3. Watch this process in htop - its name will change from `sh -c sleep 5; exec vim` to `vim-c sleep 5; exec vim`

# Cause

A "last char" index was being treated as a string length, resulting in an off-by-one error.


hishamhm:
@ScoreUnder thank you very much for both the fix and the easily-reproducible test case! Merged!

PR #815, opened 2018-07-16
Make the STARTTIME column display correctly on OpenBSD

multiplexd:
The STARTTIME column is not displayed on OpenBSD at present, though this appears to be due to the `starttime_show` field of the `struct _Process` not being suitably initialised. I have implemented this using the Solaris-specific code as reference.


hishamhm:
Thank you for the patch! Could you fix the indentation to match the rest of the code? (it looks like a Tab slipped in there)

Looking forward to get this merged, thanks! :)

multiplexd:
Indeed, I had my editor misconfigured when I was working on the code, it seems.

hishamhm:
Merged, thank you!

PR #817, opened 2018-07-17
Fix CPU meters

kremlin-:
Introduction of CP_SPIN sched state broke hard-coded state indexes
resulting in the meters incorrectly reporting bogus intr data instead of
CPU usage. Change hardcoded values to sched.h macros.

### DIFF
--- a/openbsd/Platform.c
+++ b/openbsd/Platform.c
@@ -220,7 +220,7 @@ double Platform_setCPUValues(Meter* this, int cpu) {
    percentages(CPUSTATES, diff_v, new_v,
          (int64_t *)old_v[cpu-1], scratch_v);
 
-   for (i = 0; i < CPUSTATES; i++) {
+   for (i = 0; i < CPUSTATES - 1; i++) {
       old_v[cpu-1][i] = new_v[i];
       v[i] = diff_v[i] / 10.;
    }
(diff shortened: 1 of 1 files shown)


kremlin-:
Also fix a zero index bug

hishamhm:
Thank you for looking into this — help with ports is definitely appreciated! I was reviewing that code of Platform_setCPUValues and ended up wondering: are the color bars returning values with their correct meanings on OpenBSD? When trying to follow the code, it seems it gets the vector of CPU states [straight from the sysctl](https://github.com/kremlin-/htop/blob/c005ffc3d70c14189952fa1041bab04462417308/openbsd/Platform.c#L213) and that then [goes into the this->values vector](https://github.com/kremlin-/htop/blob/c005ffc3d70c14189952fa1041bab04462417308/openbsd/Platform.c#L225) of the Meter object, which for a CPUMeter, has [this sequence](https://github.com/hishamhm/htop/blob/master/CPUMeter.c#L23-L30) (clearly based on Linux). 

So this left me wondering if the numbers match the names: on OpenBSD is the first (blue) part of the bar really showing the numbers for NICE, the second (green) bar the number for NORMAL, the third (red) the number for KERNEL, etc? (In other words, what are the names/values for those `CP_*` enum values on OpenBSD?) If they're not in the right order, then that `v` vector might need to be reordered for calculation/display...

natoscott:
Merged here: htop-dev/htop@5a17cee

PR #818, opened 2018-07-17
Update OpenBSD maximum PID

multiplexd:
From the commit message:

> The htop source code correctly states that the maximum PID number in the OpenBSD kernel is fixed in `sys/sys/proc.h`, however this was updated in revision 1.215 (two years ago!) from 32766 to 99999.


natoscott:
Merged here: htop-dev/htop@e9947ac

PR #819, opened 2018-07-17
Add environment reading support for OpenBSD

multiplexd:
Uses `kvm_getenvv(3)` to read process environment; Linux platform code used as reference for the memory allocation handling.


hishamhm:
Merged, thank you!

PR #832, opened 2018-08-23
WIP: AIX support

NattyNarwhal:
Obsoletes the previous issue. Early on, lots of stdbool twiddling, then the real meat.

What works:

* Mostly everything, except:

* strace flags need to be changed?

* IBM i support is partial; CPU meters, uptime, and load averages won't work due to lack of libperfstat, functional utmp, and no load average generation

* No battery, obviously

* The CPU meter algorithm is a little sketchy looking

For the AIX experts in the audience, you'll notice I'm taking great care not to use anything like libperfstat/ODM/WLM or worse, getting kernel symbols and twiddling them directly. This is due to a desire to keep it (mostly functional as possible) under IBM i's PASE. The most I use is an undocumented syscall in public headers to get the load average, and even then, its ifdefed out (but harmless anyways) under PASE.

A function from OpenBSD was lifted, but it seems this is OK legally speaking, seeing as the OpenBSD support in htop uses code borrowed from OpenBSD?


NattyNarwhal:
Warmer: memory meter works, (though a cosmetic glitch in bar mode?) libperfstat is used when applicable, (won't be used under i, so things there will basically be no-ops) and the swap meter. CPU is the next thing to work on, and that looks a bit hairy.

Screenshot

NattyNarwhal:
all the basic features should be working now. the memory usage meters don't seem 100% right to me, however, so I might continue tweaking those values. Feel free to look at this patch now.

NattyNarwhal:
Merge conflict resolved, Would appreciate review.

hishamhm:
Thank you for the contribution! This is a big one, I'll need to set some time aside to give it a proper review. but I will definitely get to it.

NattyNarwhal:
I just pushed something that also makes it work on IBM i's AIX syscall emulator, albeit with some functionality downgrades due to limitations there. I've been using htop with this PR on an AIX system however, and it mostly seems to work fine over the months I've been using it. The data sampling for CPU/memory isn't perfect, but it works.

Let me know about mergeability when you have the time.

Meter.c
```diff
@@ -383,7 +383,16 @@ static void GraphMeterMode_draw(Meter* this, int x, int y, int w) {
    gettimeofday(&now, NULL);
    if (!timercmp(&now, &(data->time), <)) {
       struct timeval delay = { .tv_sec = (int)(CRT_delay/10), .tv_usec = (CRT_delay-((int)(CRT_delay/10)*10)) * 100000 };
+#ifndef timeradd
```
kadler: I'd probably just hoist this code out in to a `timeradd` function and use `#ifndef timeradd` around its definition.

NattyNarwhal: I did try making the macro a function (and/or placing it in a header, but it somehow fouled up the automated header generator.

bmhm:
Can't wait for this to be merged

aix/AixProcess.h
```diff
@@ -0,0 +1,55 @@
...
+   // Add platform-specific fields here, with ids >= 100
+   WPAR_ID = 100,
+   LAST_PROCESSFIELD = 101,
+} AixProcessField;
+
+typedef struct AixProcess_ {
+   Process super;
+   int kernel;
+   cid_t cid; // WPAR ID
+   unsigned long long int stime, utime; // System/User time (stored sep for calculations)
```
kadler: I don't recommend defining multiple variables on the same line. Especially in a structure.

aix/AixProcessList.c
```diff
@@ -0,0 +1,296 @@
...
+#ifndef Process_isKernelThread
+#define Process_isKernelThread(_process) (_process->kernel == 1)
+#endif
+
+#ifndef Process_isUserlandThread
+// XXX
+#define Process_isUserlandThread(_process) (false)
+#endif
+
+}*/
```
kadler: More commented code from the header file that could be deleted.

aix/AixProcessList.h
```diff
@@ -0,0 +1,56 @@
...
+htop - AixProcessList.h
+(C) 2014 Hisham H. Muhammad
+(C) 2018 Calvin Buckley
+Released under the GNU GPL, see the COPYING file
+in the source distribution for its full text.
+*/
+
+#ifndef __PASE__
+#else
+#endif
```
kadler: nop

Meter.c
kadler: I suppose since this is the only usage in the code, there's not much point separating it out in to a function/macro. My only suggestion then would be to reverse the check to `#ifdef timeradd` - keep the normal path first.

StringUtils.c
```diff
@@ -26,6 +27,27 @@ in the source distribution for its full text.
  * at compile time (e.g. when they are immutable string literals). :)
  */
 
+#ifndef HAVE_STRCASESTR
+// OpenBSD strcasestr
```
kadler: If this came from OpenBSD, you need to include the BSD license it was under somewhere in the code.

aix/AixProcess.c
```diff
@@ -0,0 +1,243 @@
...
+#ifndef Process_isKernelThread
+#define Process_isKernelThread(_process) (_process->kernel == 1)
+#endif
+
+#ifndef Process_isUserlandThread
+// XXX
+#define Process_isUserlandThread(_process) (false)
+#endif
+
+}*/
```
kadler: This looks like it was from the header file and should be removed from the code.

aix/AixProcess.c
```diff
@@ -0,0 +1,243 @@
...
+   AixProcess *p1, *p2;
+   Settings *settings = ((Process*)v1)->settings;
+   if (settings->direction == 1) {
+      p1 = (AixProcess*)v1;
+      p2 = (AixProcess*)v2;
+   } else {
+      p2 = (AixProcess*)v1;
+      p1 = (AixProcess*)v2;
+   }
+   switch ((int) settings->sortKey) {
```
kadler: Why the cast?

CheckItem.h
```diff
@@ -11,6 +11,8 @@ in the source distribution for its full text.
 
 #include "Object.h"
 
+#include <stdbool.h>
+
```
kadler: This looks like they current code is non-standardly using C99 `bool` without include stdbool.h. This could probably be done as a separate PR just to shrink this PR a bit.

Process.h
```diff
@@ -29,6 +29,7 @@ in the source distribution for its full text.
 #include "Object.h"
 
 #include <sys/types.h>
+#include <stdbool.h>
```
kadler: Same as in CheckItem.h

htop.h
```diff
@@ -9,6 +9,9 @@ Released under the GNU GPL, see the COPYING file
 in the source distribution for its full text.
 */
 
+#ifdef HAVE_GETOPT_H
+#endif
+
```
kadler: nop?

kadler (changes requested):
I would rather see AIX support get merged first before adding support for PASE. I think there's enough differences between the two that having separate files makes sense. I'd love to see htop have job id and other IBM i-specific fields, but that will get way too complicated to support mixed in with AIX.

aix/AixProcessList.c
NattyNarwhal: This is actually how htop's buid system works - the headers have things from the comments transcluded into them, from what I understand (!).

NattyNarwhal:
Some of the suggestions made. A lot of this is quirks of htop itself and its build system... and also me from two years ago being pretty dumb.

bmhm:
Hi. 
I tested the PR about two weeks ago. Had some problems with the default colour scheme, but the overview (main page) worked just fine.

If you need me to test sth on AIX 7.2, just tell me.

aix/AixProcessList.c
kadler: Ah, interesting!

PR #838, opened 2018-09-16
*bsd: mark unreachable code with comment

zebastian:
This may be undesired and the code should be revisited by the maintainers of these two occurances.

### DIFF
--- a/openbsd/OpenBSDProcess.c
+++ b/openbsd/OpenBSDProcess.c
@@ -209,7 +209,7 @@ void OpenBSDProcess_writeField(Process* this, RichString* str, ProcessField fiel
       Process_writeField(this, str, field);
       return;
    }
-   RichString_append(str, attr, buffer);
+   // RichString_append(str, attr, buffer); // <!-- unreachable
 }
 
 long OpenBSDProcess_compare(const void* v1, const void* v2) {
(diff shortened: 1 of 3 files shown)


Explorer09:
Leaving non-trivial unreachable code there is surely a sign of bugs. It's good to let *BSD port developers look at those. Or do you have the time to bother and fix them?

zebastian:
both instances seem to be copy paste issues, since there are very similar codes in the other 
```[vendor]/[Vendor]Process.c```
files. 
Both dead code is IMHO harmless and can be savely deleted without replacement.
However I am not totally sure about it, so i rather would make that up to the maintainers.

File: openbsd/OpenBSDProcess.c [hunk 1]
```diff
@@ -209,7 +209,7 @@ void OpenBSDProcess_writeField(Process* this, RichString* str, ProcessField fiel
       Process_writeField(this, str, field);
       return;
    }
-   RichString_append(str, attr, buffer);
+   // RichString_append(str, attr, buffer); // <!-- unreachable
```
hishamhm: for this one, it looks like the entire `OpenBSDProcess_writeField function` is there just as a placeholder to make it easier to add OpenBSD-specific fields in the future. The whole function could be removed and just replaced with `Process_writeField` in the definition of `OpenBSDProcess_class`. Did you add these comments to silence warnings in some specific platform?

Issue #841, opened 2018-10-07
ST_UID column needs to be 5 chars wide for UIDs like nobody=65534


fasterit:
The ST_UID column is laid out to be 4 chars wide and UIDs > 9999 then lead to misalignment of the columns.

Downstream bug report in Debian:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=910492
(includes screenshot)

PR coming

Explorer09:
Can we have a UID column that dynamcially adjusts its own width, just like how PID column does?

PR #842, opened 2018-10-07
Widen ST_UID (UID) column to 5 chars to allow UIDs > 9999 without bre…

fasterit:
Widen ST_UID (UID) column to 5 chars to allow UIDs > 9999 without breaking alignment
Issue Github #841, Debian bug #910492


natoscott:
Merged here: https://github.com/htop-dev/htop/commit/579995c7c2385b844d81b2508710dcbbe9b0d64f

PR #843, opened 2018-10-09
Add PSS (proportional set size), Swap and SwapPSS calculation on Linux

ntninja:
I did not write this code and took a patch for this from https://github.com/linvinus/htop-mod. Please tell me about any improvements required.

While I can attest PSS to be a big improvement over RSS, there are some pain points about this:

  * `/proc/<pid>/smaps` is only readable by the (effective?) user of the given process. ~Since PSS calculation requires all memory mapping tables to be available for accurate calculation, this makes the PSS value somewhat useless `htop` runs as root.~ Actually this is not true making PSS data somewhat usable, but incomplete for users other than the current user, unless you're root. :slightly_frowning_face: 
  * Greatly improved on Linux 4.14+ thanks to @himikof's suggestion! (Crunching the `smaps` data noticeably slows down `htop` – far from unusable, but slower.)

Original code was written by *Craig M. Brandenburg* for htop 1.0.2
Many performance improvements by GitHub user *linvinus*, ported to htop 2.0.2

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -89,6 +89,7 @@ typedef struct LinuxProcessList_ {
    
    CPUData* cpus;
    TtyDriver* ttyDrivers;
+   bool haveSmapsRollup;
    
    #ifdef HAVE_DELAYACCT
    struct nl_sock *netlink_socket;
@@ -240,8 +241,17 @@ ProcessList* ProcessList_new(UsersTable* usersTable, Hashtable* pidWhiteList, ui
    LinuxProcessList_initNetlinkSocket(this);
    #endif
 
+   // Check for /proc/*/smaps_rollup availability (improves smaps parsing speed, Linux 4.14+)
+   FILE* file = fopen(PROCDIR "/self/smaps_rollup", "r");
+   if(file != NULL) {
+      this->haveSmapsRollup = true;
+      fclose(file);
+   } else {
+      this->haveSmapsRollup = false;
+   }
+
    // Update CPU count:
-   FILE* file = fopen(PROCSTATFILE, "r");
+   file = fopen(PROCSTATFILE, "r");
    if (file == NULL) {
       CRT_fatalError("Cannot open " PROCSTATFILE);
    }
@@ -486,7 +496,7 @@ static bool LinuxProcessList_readStatmFile(LinuxProcess* process, const char* di
    return (errno == 0);
 }
 
-static bool LinuxProcessList_readSmapsFile(LinuxProcess* process, const char* dirname, const char* name) {
+static bool LinuxProcessList_readSmapsFile(LinuxProcess* process, const char* dirname, const char* name, bool haveSmapsRollup) {
    //http://elixir.free-electrons.com/linux/v4.10/source/fs/proc/task_mmu.c#L719
    //kernel will return data in chunks of size PAGE_SIZE or less.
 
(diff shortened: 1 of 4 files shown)


himikof:
Maybe the code could be extended to try [smaps_rollup](https://marc.info/?l=linux-mm&m=150250453624272&w=2) if available (in kernel since 4.14)? The format is exactly the same, just less data to parse.

alexander255:
@himikof: Thanks for the suggestion! I've implemented your suggestion and it does give quite a performance boost. Still not as quite snappy as when not reading any smaps file, but it stops the UI from feeling somewhat sluggish while PSS reporting is enabled.

hishamhm:
Thanks for the patch! I still need to take some time to give it a spin, but on first glance my initial comment is that we'll probably not want to enable PSS by default, given the performance concerns.

linux/LinuxProcessList.c
```diff
@@ -486,6 +496,62 @@ static bool LinuxProcessList_readStatmFile(LinuxProcess* process, const char* di
...
+      return false;
+
+   process->m_pss   = 0;
+   process->m_swap  = 0;
+   process->m_psswp = 0;
+
+   while ( ( nread =  read(fd,buffer, sizeof(buffer)) ) > 0 ){
+        start = (char *)&buffer;
+        end   = start + nread;
+        do{//parse 4k block
```
hishamhm: I am afraid this loop won't read correctly the data from pages that fall in between two 4096-byte reads:

```
$ cat /proc/3304/task/3314/smaps | head -c 4096 | tail -n 10 
VmFlags: mr mw me 
2ec0a4f2b000-2ec0a4f3b000 r-xp 00000000 00:00 0 
Size:                 64 kB
Rss:                  64 kB
Pss:                  64 kB
Shared_Clean:          0 kB
Shared_Dirty:          0 kB
Private_Clean:         0 kB
Private_Dirty:        64 kB
Referenced:      %
```

For the above entry, it will read the Pss for this but not Swap and SwapPss, right?

I think we could get around the whole issue if we just add this feature for systems that have `smaps_rollup` available. Reading only the rollup version will simplify the patch and avoid the performance concerns.

linux/LinuxProcessList.c
```diff
@@ -486,6 +496,62 @@ static bool LinuxProcessList_readStatmFile(LinuxProcess* process, const char* di
    return (errno == 0);
 }
 
+static bool LinuxProcessList_readSmapsFile(LinuxProcess* process, const char* dirname, const char* name, bool haveSmapsRollup) {
+   //http://elixir.free-electrons.com/linux/v4.10/source/fs/proc/task_mmu.c#L719
+   //kernel will return data in chunks of size PAGE_SIZE or less.
```
alexander255: @hishamhm: True and well analyzed! However, such reads will never happen for regular `smaps` because data is returned in chunks of PAGE_SIZE. For `smaps_rollup`, as you have correctly pointed out, this discussion is irrelevant anyways.
Regarding the verification of this claim: The linked code does appear to write out each result entry separately and when I run `strace cat /proc/$$/smaps` each read returns as many entries as fit into a 4k-page, without any “half-entries”. (For the most part we have to trust the previous author's judgment here. Fortunately however we do not need to worry about kernel changes, since we're only really talking about Linux ~2.6.26 up to 2.13 here).

alexander255:
@hishamhm: I have already explained why this is not a problem in the comment above. Mind considering a merge again?

alexander255:
@hishamhm: Ping

natoscott:
Merged here: htop-dev/htop@f9625ca

PR #847, opened 2018-10-26
Change countCPUsFromZero to true 

noobmanbr:


hishamhm:
Thanks for this PR, but it makes an undesired change of the default behavior. I'll be keeping this off by default as is. Thanks.

PR #848, opened 2018-10-31
support threads on OSX

robaho:
It's a bit rough, because the OSX thread ID is not a PID, but it does work.

### DIFF
--- a/darwin/DarwinProcess.c
+++ b/darwin/DarwinProcess.c
@@ -365,14 +365,17 @@ void DarwinProcess_scanThreads(DarwinProcess *dp,DarwinProcessList *dpl) {
       if (ret == KERN_SUCCESS) {
          uint64_t tid = 0 - thread_list[i];
          thread_extended_info_t eti = (thread_extended_info_t) thinfo;
-         if(strlen(eti->pth_name)==0 || dpl->super.settings->hideThreads)
+         if(dpl->super.settings->hideThreads)
              continue;
          bool preExisting;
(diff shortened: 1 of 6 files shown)


Kronuz:
Thank you! This is working as advertised! (macOS Mojave)

It does, however, need `sudo htop` so threads are shown.

robaho:
Yes, in order to remove the sudo requirement, there is another OSX API that needs to be used, and the application needs to be signed - possible but a lot more work in the build/packaging phase

joshuawarner32:
Hi robaho! This is exciting, thanks! Do you know what that other API you mentioned is, by chance (the one that requires signing)?

robaho:
@joshuawarner32 not really a different API, but a different process. see http://os-tres.net/blog/2010/02/17/mac-os-x-and-task-for-pid-mach-call/

It is probably easier to set the setuid bit on the executable like 'top' does.

joshuawarner32:
Oh, hmm... based on my (brief, incomplete) testing, it looks like calling `task_for_pid` requires SIP to be disabled, even when running as root.  I didn't try a setuid bit, but I'd be a little surprised if that had different behavior.

Signing could probably work - but having people install a self-signed code signing cert sounds kinda sketchy.

robaho:
The code uses task_for_pid now and I don’t have SIP disabled and it works for me.

joshuawarner32:
Ah, interesting.  I was testing against Google Chrome, which seems to have some additional anti-debugging protections enabled, and does require disabling SIP before listing the threads.  When I try other processes on my system, they generally work fine.

robaho:
Yea, it certainly isn’t easy to find official docs on this stuff, probably intentionally.

Issue #601, opened 2017-02-20
Handle less search navigation shortcuts after hitting enter in search bar


aalexgabi:
What do you think about handling less search navigation shortcuts after hitting enter in search bar?
* n - go to next search result
* Shift + n - go to previous search result
* Shift + P - go to the begining

Explorer09:
n and Shift+n will collide with "Append n or N to the search query string". I don't think that's possible.

aalexgabi:
@Explorer09 Not AFTER hitting enter in the search bar:
* Press F3
* Type search query
* Press enter (now first line should be selected)
* Press n or Shift + n to navigate to other processes matching the search query

Explorer09:
@aalexgabi 
Even with that, `top` has allocated 'n' and 'N' for other purposes now ('n'= set maximum number of processes to display in list; 'N'= sort process list by PID). Introducing them in htop can cause user confusion.

Explorer09:
@aalexgabi
By the way, `top` uses '&' for this "locate next occurrence of search query" purpose.

PR #850, opened 2018-11-03
Added MainPanel actions n and N for find next and find prev

syrrim:
this was discussed in issue #601:

- I didn't add an action for P because this has since been used

- I don't think concerns regarding overlap with top commands are overly important, because the relevant top commands haven't been added in the meanwhile

Besides that, I think that these are very standard key combos, being used in a variety of unix tools as well as less.


natoscott:
Merged here: htop-dev/htop@46ab1aa

PR #857, opened 2018-11-11
Don't follow process when selecting non-process-specific options

### LINKED ISSUE #856: "Show processes of" user stops working if user logs out.
Steps to reproduce:

1) Log in as two different users, call them Alice and Bob.
2) Alice runs htop.
3) Alice presses "u" to bring up "Show processes of" menu.
4) Alice selects Bob.
5) Alice can now see Bob's processes.
6) Bob logs out, ending all of Bob's process.
7) Alice's htop now displays no processes.
8) Alice presses "u" to bring up "Show processes of" menu.
9) Alice selects any user (or all users).
10) Alice's htop continues to show no processes.

10 is the bug.  I would expect Alice to see the processes of the just selected user (or users).

For fun:

11) Bob logs back in.
12) Alice now sees Bob's processes, even though in step 9, Alice selected some other user.

Tested with htop 2.20 on Ubuntu 18.04.  (The problem is also in htop 2.1.0, which is the version in the Ubuntu 18.04 repositories.)

wangqr:
Currently `Action_pickFromVector()` will follow the selected process and check if it remains the same after an option is selected. While this is useful when performing process-specific actions (so you don't kill another process by accident), this behavior should be disabled when selecting non-process-specific options, such as selecting sort order or selecting user filter.

This PR added a new parameter, `bool followProcess` to `Action_pickFromVector()`, and only follows the process when it is set to `true`.

This PR will fix #856

### DIFF
--- a/Action.c
+++ b/Action.c
@@ -62,7 +62,7 @@ typedef struct State_ {
 
 }*/
 
-Object* Action_pickFromVector(State* st, Panel* list, int x) {
+Object* Action_pickFromVector(State* st, Panel* list, int x, bool followProcess) {
    Panel* panel = st->panel;
    Header* header = st->header;
    Settings* settings = st->settings;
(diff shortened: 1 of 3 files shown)


natoscott:
Merged here: htop-dev/htop@e7d2f93

PR #867, opened 2018-12-15
Fix warning about "major" macro with glibc 2.27

wataash:
Fixes:

    gcc -DHAVE_CONFIG_H -I. -I..  -DNDEBUG  -pedantic -Wall -Wextra -std=c99 -D_XOPEN_SOURCE_EXTENDED -DSYSCONFDIR=\"/home/wsh/usr/opt/htop/etc\" -I"../linux" -rdynamic -D_GNU_SOURCE -D_DEFAULT_SOURCE -I/usr/include/ncursesw -ggdb3 -O0 -MT Process.o -MD -MP -MF $depbase.Tpo -c -o Process.o ../Process.c &&\
    mv -f $depbase.Tpo $depbase.Po
    ../Process.c: In function ‘Process_writeField’:
    ../Process.c:473:13: warning: In the GNU C Library, "major" is defined
     by <sys/sysmacros.h>. For historical compatibility, it is
     currently defined by <sys/types.h> as well, but we plan to
     remove this soon. To use "major", include <sys/sysmacros.h>
     directly. If you did not intend to use a system-defined macro
     "major", you should undefine it after including <sys/types.h>.
        case TTY_NR: xSnprintf(buffer, n, "%3u:%3u ", major(this->tty_nr), minor(this->tty_nr)); break;
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

In the previous configure.ac, MAJOR_IN_SYSMACROS was not defined even if
"major" is defined in <sys/sysmacros.h>.

Code snnipet is copied from:
https://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=blob;f=lib/autoconf/headers.m4;hb=e17a30e987d7ee695fb4294a82d987ec3dc9b974

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -64,15 +64,22 @@ AC_CHECK_HEADERS([stdlib.h string.h strings.h sys/param.h sys/time.h unistd.h],[
 ])
 AC_CHECK_HEADERS([execinfo.h],[:],[:])
 
-AC_HEADER_MAJOR
 dnl glibc 2.25 deprecates 'major' and 'minor' in <sys/types.h> and requires to
-dnl include <sys/sysmacros.h>. However the logic in AC_HEADER_MAJOR has not yet
-dnl been updated in Autoconf 2.69, so use a workaround:
-m4_version_prereq([2.70], [],
-[if test "x$ac_cv_header_sys_mkdev_h" = xno; then
-   AC_CHECK_HEADER(sys/sysmacros.h, [AC_DEFINE(MAJOR_IN_SYSMACROS, 1,
-      [Define to 1 if `major', `minor', and `makedev' are declared in <sys/sysmacros.h>.])])
-fi])
+dnl include <sys/sysmacros.h>. Autoconf<=2.69 doesn't care about it.
+AC_DEFUN([AC_HEADER_MAJOR_270], dnl from autoconf lib/autoconf/headers.m4 @e17a30e987d
+[AC_CHECK_HEADERS_ONCE([sys/types.h])
+AC_CHECK_HEADER(sys/mkdev.h,
+    [AC_DEFINE(MAJOR_IN_MKDEV, 1,
+         [Define to 1 if `major', `minor', and `makedev' are
+          declared in <mkdev.h>.])])
+if test $ac_cv_header_sys_mkdev_h = no; then
+  AC_CHECK_HEADER(sys/sysmacros.h,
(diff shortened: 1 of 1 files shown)


Explorer09:
@wataash Could you please answer, how is your proposed change different from the workaround I have presented (I'm the one who made the original pull request)?
I wish the workaround code be small, just enough to address the problem with Autoconf at hand. Pulling the Autoconf code and embedding it here would be an overkill, unless you can justify why you must do so.

Explorer09:
@wataash Ok, I've fixed the logic. Try #872 if you can.

wataash:
@Explorer09
Thank you for reviewing my changes.
I'll check it in a few days...

wataash:
> how is your proposed change different from the workaround I have presented (I'm the one who made the original pull request)?

Even after merging #872, there're slight differences, for example:

- If `<sys/types.h>`, `<sys/mkdev.h>` and `<sys/sysmacros.h>` all defines the `major` macro, autoconf 2.69 defines `MAJOR_IN_SYSMACROS` and 2.70 does `MAJOR_IN_MKDEV`.
- If `<sys/types.h>` doesn't exists and `<sys/mkdev.h>`, autoconf 2.69 defines `MAJOR_IN_MKDEV` and 2.70 does nothing.

> I wish the workaround code be small,

Yes, quite agree with it.

> Pulling the Autoconf code and embedding it here would be an overkill, unless you can justify why you must do so.

I don't think so. Code in Autoconf should be well tested and considered to be less buggy. The reason I did it was to reduce differences of behaviors between Autoconf <=2.69 and 2.70.

However, #872 fixes this issue and much simpler than mine, and I believe there won't arise other issues, let me close my pull request.

Thank you!

Explorer09:
@wataash Just some explanations:

> If `<sys/types.h>`, `<sys/mkdev.h>` and `<sys/sysmacros.h>` all defines (sic) the `major` macro, autoconf 2.69 defines `MAJOR_IN_SYSMACROS` and 2.70 does `MAJOR_IN_MKDEV`.

The expectation is that including either header would work if both headers define `major`. If the definitions differ, it would be the problem of the builder's platform and not our business (there's no proper "precedence" of which header should be chosen first; the order of checking the headers is arbitrary).

PR #868, opened 2018-12-16
Fix CPU usage on OpenBSD

motet-a:
The current OpenBSD-specific CPU usage code is broken. The `cpu`
parameter of `Platform_setCPUValues` is an integer in the interval
[0, cpuCount], not [0, cpuCount-1]: Actual CPUs are numbered from
1, the “zero” CPU is a “virtual” one which represents the average
of actual CPUs (I guess it’s inherited from Linux’s `/proc/stats`).
This off-by-one error leads to random crashes.

Moreover, the displayed CPU usage is more detailed with system,
user and nice times.

I made the OpenBSD CPU code more similar to the Linux CPU code,
removing a few old bits from OpenBSD’s top(1). I think it will be
easier to understand, maintain and evolve.

I’d love some feedback from experienced OpenBSD people.

### DIFF
--- a/openbsd/OpenBSDProcessList.c
+++ b/openbsd/OpenBSDProcessList.c
@@ -6,6 +6,7 @@ Released under the GNU GPL, see the COPYING file
 in the source distribution for its full text.
 */
 
+#include "CRT.h"
 #include "ProcessList.h"
 #include "OpenBSDProcessList.h"
 #include "OpenBSDProcess.h"
@@ -17,6 +18,7 @@ in the source distribution for its full text.
 #include <sys/param.h>
 #include <sys/proc.h>
 #include <sys/resource.h>
+#include <sys/sched.h>
 #include <sys/sysctl.h>
 #include <sys/types.h>
 #include <sys/user.h>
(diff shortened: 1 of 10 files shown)


openbsd/OpenBSDProcessList.c
```diff
@@ -290,14 +306,99 @@ void ProcessList_goThroughEntries(ProcessList* this) {
...
+      totalTime += times[i];
+   }
+
+   unsigned long long sysAllTime = times[CP_INTR] + times[CP_SYS];
+
+   // XXX Not sure if CP_SPIN should be added to sysAllTime.
+   // See https://github.com/openbsd/src/commit/531d8034253fb82282f0f353c086e9ad827e031c
+   #ifdef CP_SPIN
+   sysAllTime += times[CP_SPIN];
+   #endif
```
motet-a: I used `#ifdef CP_SPIN` everywhere because `CP_SPIN` is pretty much new and not available on 6.3.

openbsd/Platform.c
```diff
@@ -201,43 +151,37 @@ void Platform_getLoadAverage(double* one, double* five, double* fifteen) {
 
 int Platform_getMaxPid() {
    // this is hard-coded in sys/sys/proc.h - no sysctl exists
-   return 32766;
+   return 32766; // XXX Seems changed to 99999, what about using PID_MAX?
```
motet-a: (I should fix this in another PR)

natoscott:
Merged here: htop-dev/htop@78f4d06

Issue #733, opened 2018-02-08
version 2.1.0 does not run in a small terminal


danthem:
I use i3-wm on Arch and I like to run htop in a small terminal in the lower right corner of my screen. This worked fine up until I updated to version 2.1.0, since then if the terminal window is too small htop will not launch.

```
[danthem@machine ~]$ htop
htop: Success 
```
making the window bigger allows htop to run but I'd like to know if this was an intended change or a bug? I'd rather keep the window small.

Terminal is urxvt v9.22.

Explorer09:
How small for the terminal window will htop exit as you described? Note that exiting with `errno` equals Success is definitely a bug. We just want you to help us track which change caused it (in case you know how to use `git bisect` to narrow down the changes).

fasterit:
Downstream bug report at Debian BTS: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=889693
Reproducible in plain xterm (Debian Stretch).

danthem:
@Explorer09 
Looks like it's mostly about the width of the window rather than the height. 

This window is too small and results in the strange Success-exit:
```
[danthem@machine ~]$ htop
htop: Success
[danthem@machine ~]$ stty size
10 38
[danthem@machine ~]$ xwininfo -id 0xe0000a
[...]
Width: 391
Height: 201
[...]

```

However it runs fine at:
```
[danthem@machine ~]$ stty size
10 40
[danthem@machine ~]$ xwininfo -id 0xe0000a
[...]
Width: 404
Height: 201
[...]
```

ginggs:
The problem lies here:
https://github.com/hishamhm/htop/blob/f37a050d3d0c6928b09d9b19e1283a695c52ccc7/Meter.c#L290
reverting this line to:
`snprintf(bar, w + 1, "%*s", w, buffer);`
allows htop to run in a small terminal again.

PR #869, opened 2018-12-18
Fix numbers larger than 100 terabytes

adrien1018:
This is an issue similar to #733. If a process has `RD_CHAR` or `WR_CHAR` larger than 100 TB, and the columns are displayed, htop will terminate and print `htop: Success`.

The situation can be reproduced by running something that calls `read`/`write` a lot for a few hours, such as `cat /dev/zero > /dev/null`.

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -228,7 +228,7 @@ void Process_humanNumber(RichString* str, unsigned long number, bool coloring) {
    if(number >= (10 * ONE_DECIMAL_M)) {
       #ifdef __LP64__
       if(number >= (100 * ONE_DECIMAL_G)) {
-         len = snprintf(buffer, 10, "%4ldT ", number / ONE_G);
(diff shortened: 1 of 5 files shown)


Explorer09:
Can we have a solution that's more future-proof? Even if we fix the 100 TB problem now, there is still a potential of snprintf() overflow on larger integers, e.g. petabytes or exabytes (2^64 = 16 exa).

And according to the reproduce case, this should a local denial-of-service security vulnerability.

adrien1018:
OK, now it can deal with any `RD_CHAR` inside 64-bit range. It can also deal with I/O rate up to 10PB/sec now.

(There are still some theoretical situations that can make snprintf() overflow, e.g. memory larger than 10PB or jobs running more than 1140 years. Should we also deal with these extreme cases?)

natoscott:
Merged here: htop-dev/htop@00d333c

PR #871, opened 2018-12-24
Fix memory statistics display on FreeBSD/powerpc

ghost:
Use the appropriate types when calling sysctl().

Currently, `unsigned long long int` is used for all sizes and on
FreeBSD/powerpc this causes all sysctl() calls in scanMemoryInfo()
to fail as they are actually of different sizes on powerpc, where
(sizeof(unsigned long long int), sizeof(u_long)) == (8, 4)
vs (8, 8) on amd64.  This results in bogus memory sizes being
reported by htop.

Signed-off-by: Tobias Kortkamp <tobik@FreeBSD.org>

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -228,7 +228,7 @@ void Process_humanNumber(RichString* str, unsigned long number, bool coloring) {
    if(number >= (10 * ONE_DECIMAL_M)) {
       #ifdef __LP64__
       if(number >= (100 * ONE_DECIMAL_G)) {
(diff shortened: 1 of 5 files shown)


t6:
To illustrate:
Before patching: https://people.freebsd.org/~tobik/htop-powerpc-bug/before.png
After patching: https://people.freebsd.org/~tobik/htop-powerpc-bug/after.png

natoscott:
Merged here: htop-dev/htop@ed2b779

PR #746, opened 2018-02-24
Fix build failure ('major' undefined) in glibc 2.28.

Explorer09:
glibc 2.28 no longer defines 'major' and 'minor' in <sys/types.h> and
requires us to include <sys/sysmacros.h>. (glibc 2.25 starts
deprecating the macros in <sys/types.h>.) Now do include the latter if
found on the system.

At the moment, let's also utilize AC_HEADER_MAJOR in configure script.
However as Autoconf 2.69 has not yet updated the AC_HEADER_MAJOR macro
to reflect the glibc change [1], we need a workaround code.

Fixes #663. Supersedes pull request #729.

Reference:
[1] https://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=commit;h=e17a30e987d7ee695fb4294a82d987ec3dc9b974

PR #872, opened 2018-12-30
Fix configure 'major' check causing <sys/sysmacros.h> to miss.

Explorer09:
A logic mistake in pull request #746 causes <sys/sysmacro.h> to be
*not* included when AC_HEADER_MAJOR (before autoconf-2.70) finds
'major' in <sys/types.h>. Though this would still build htop, it would
still bring deprecation warning in systems using glibc 2.25-2.27. Fix
the logic and suppress the warning.

Also, include config.h in Process.c for the sake of strengthening the
code.

(This should supersede pull request #867. Thanks @wataash for discovering this.)

### DIFF
--- a/Process.c
+++ b/Process.c
@@ -8,6 +8,8 @@ in the source distribution for its full text.
 #include "Process.h"
 #include "Settings.h"
 
+#include "config.h"
+
 #include "CRT.h"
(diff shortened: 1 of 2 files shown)


natoscott:
Merged here: htop-dev/htop@dab4144

Issue #873, opened 2019-01-02
Unable to build on Solaris 11.4


m4r1k:
Hey there,

I've tried to build htop on Solaris 11.4 but it fails in the early make phases.

```sh
CRT.h:113:4: error: expected identifier before numeric constant
    CPU_KERNEL,
    ^
In file included from ColumnsPanel.c:13:0:
CRT.h:145:47: error: ‘LAST_COLORELEMENT’ undeclared here (not in a function); did you mean ‘LAST_COLORSCHEME’?
 extern int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT];
                                               ^~~~~~~~~~~~~~~~~
                                               LAST_COLORSCHEME
```

Follows the full log
https://pastebin.com/xt2VbccD

Did I do something wrong?

Thanks!

ninefathom:
Found the issue.  Solaris 11.4 now includes <sys/sysinfo.h> from <sys/zone.h>.  Previous Solaris versions did not, and illumos still does not.  That inclusion is causing a macro name collission with htop.  I've have a PR to fix this at some point in the next day or two.

In the mean time, please try my solaris-114_fixup branch from https://github.com/ninefathom/htop and let me know if that takes care of it.

m4r1k:
Hey ninefathom,

Thanks for the quick reply.
I've tested the branch solaris-114_fixup and make fails at the same point.

```sh
In file included from /usr/include/sys/zone.h:28:0,
                 from /usr/include/sys/corectl.h:9,
                 from /usr/include/libproc.h:36,
                 from ./solaris/Platform.h:20,
                 from ColumnsPanel.c:9:
CRT.h:113:4: error: expected identifier before numeric constant
    CPU_KERNEL,
    ^
In file included from ColumnsPanel.c:13:0:
CRT.h:145:47: error: ‘LAST_COLORELEMENT’ undeclared here (not in a function); did you mean ‘LAST_COLORSCHEME’?
 extern int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT];
                                               ^~~~~~~~~~~~~~~~~
                                               LAST_COLORSCHEME
*** Error code 1
make: Fatal error: Command failed for target `ColumnsPanel.o'
```

ninefathom:
@m4r1k Whoops.  Made a change but forgot to push - dangers of coding while half asleep.  Please do a "git pull" and try again.

m4r1k:
[...]

PR #875, opened 2019-01-03
Fix build on Solaris 11.4

ninefathom:
HEADS UP - contains changes to platform-independent code!

Solaris 11.4 made a minor change to system headers, which results in a collision between the macro CPU_KERNEL in found Solaris <sys/sysinfo.h> and the htop color CPU_KERNEL in CRT.h (see issue #873 ).

After reviewing multiple avenues for a fix, my conclusion was that the color names in CRT.h under ColorElements_ are a bit too generic, and fixing this just for Solaris would likely just leave us open to similar issues down the road.  As such, I went ahead and prefixed all of the color names with COLOR_ and updated all references to them throughout htop.

Changes tested w/ full build and run on:
 - Ubuntu 16.04 LTS (x86_64)
 - Fedora 29 (x86_64)
 - OpenBSD 6.4 (x86_64)
 - Solaris 11.4 (x86-64 and SPARC)
 - OpenIndiana Hipster


ninefathom:
Deprecated by a newer PR with additional work

Issue #865, opened 2018-12-14
Solaris port: memory graphs should report zone limits correctly


ninefathom:
Report zone RSS cap rather than system memory

_Originally posted by @papertigers in https://github.com/hishamhm/htop/pull/741#issuecomment-447405051_

ninefathom:
@papertigers - I've run into a major snag here.  While the per-zone _limits_ are available via kstat, the per-zone _usage_ is not, and instead must be gotten via getvmusage() and friends- that's how zonestatd does it, for example.  The problem?  For a zone with RSS enabled, getting the total used memory for that zone is very, very slow.  Too slow for a rapidly-refreshing list of processes.

Still looking at alternative approaches, but unless a better solution presents itself, this feature request might be DoA.

PR #880, opened 2019-01-23
Various Solaris Fixes

ninefathom:
**HEADS UP - contains changes to platform-independent code!**

### Part 1: Broken build on Solaris 11.4
Solaris 11.4 made a minor change to system headers, which results in a collision between the macro CPU_KERNEL in found Solaris <sys/sysinfo.h> and the htop color CPU_KERNEL in CRT.h (see issue #873 ).

After reviewing multiple avenues for a fix, my conclusion was that the color names in CRT.h under ColorElements_ are a bit too generic, and fixing this just for Solaris would likely just leave us open to similar issues down the road. As such, I went ahead and prefixed all of the color names with COLOR_ and updated all references to them throughout htop.

### Part 2: Memory meter under Solaris that is aware of zone limits
In issue #865 a user reported that the memory graph, while in a zone with memory limits enabled, would show system memory usage and limits, rather than those of the zone, which was misleading.

Pursuant to that, I've added a new Solaris-specific memory meter which, while in a memory-limited zone, will show the zone's memory usage with (default theme) the usual green, the zone limit with blue (the Solaris port doesn't report buffer or cached memory anyway, so no collision there), and the system usage with gray. The maximum value is the system maximum, but a user can easily see where they sit on the zone limit. As this new meter is compatible with the base one from htop, it is the default on Solaris. No htoprc changes are needed.

This is not precisely what the user in issue #865 originally requested, but it mostly fills the information void, and I think that it's the best we're going to get.

### Part 3: Fixed a long-standing bug in reporting process CPU time
Units. Incorrect unit conversion from Solaris timestruc_t to htop time, simple as that... but very tedious to track down.

### Part 4: Add a new Solaris-specific column to show data model (32- vs. 64-bit)
Ongoing push in the illumos community to move remaining 32-bit binaries to 64-bit - hopefully having this available as a data column in htop can help.

### Part 5: Add a new column to show process security flags (works on illumos only)
If the user account under which htop is running can access a given process's security flags, they will be visible as the new "PSEC" column. A = ASLR, F = forbidnullmap, N = noexecstack. "-" means the flag is not set, a blank value means the information is not available. Oracle Solaris always shows the column as blank, since process security was implemented after the schism, and done very differently on the two platforms.

### Part 6: New Solaris-specific TTY column
Solaris's (and illumos's) dev_t is larger than the (unsigned int) allocated for it in htop's Process.c . After some testing, changing the size in platform-independent code introduced a host of other issues. As such, a new custom SOLTTY_NR column for Solaris / illumos seemed the best option.


ninefathom:
Re-did this PR with a clean branch from hishamhm/htop:master, and pulled in relevant changes one at a time.  It looks like GitHub was confused by prior squashed Solaris commits, and was including lots of stuff in the earlier PR that had already been incorporated upstream.

m4r1k:
Emh, how come 4 months after this Solaris specific series has no been merged yet nor feedback was provided ... As of now, htop on Solaris is broken (wrong cpu utilization report) and on latest version it won't even compile without this patch
@ninefathom

ninefathom:
> Emh, how come 4 months after this Solaris specific series has no been merged yet nor feedback was provided ... As of now, htop on Solaris is broken (wrong cpu utilization report) and on latest version it won't even compile without this patch
> @ninefathom

Not sure, @m4r1k .  Our friend @hishamhm seems to be alive and well - perhaps he is taking a break from htop to work on other projects?  As a rule, I try not to be too demanding regarding FOSS.

Mno-hime:
Great achievement @ninefathom! I tested this on OpenIndiana, I especially like the correct memory usage in zone. Ported your patch on top of our packaged htop 2.2.0, should anyone wanted to try out: https://github.com/Mno-hime/oi-userland/tree/htop_htop_pr_880.

m4r1k:
@ninefathom FOSS is not even minimally related to the issue here. It's about when a project fully relies on one person. This can happen in FOSS as much as in closed source projects.

When the project is also widely used, it gets ugly.

ninefathom:
PR has been open for nearly a year with no maintainer input.  As a result, current Solaris code is now being maintained at https://github.com/ninefathom/htop/tree/solaris-stable .  Will re-create an updated PR once @hishamhm is accepting them again.

PR #881, opened 2019-01-26
chore: add clarification to the docs

ukiahsmith:
This give a bit detail as to why the INSTALL file isn't listed in the repo.

Also, the link in this line is 404, because the file is not yet generated. We should unlink this to prevent confusion.


natoscott:
Merged here: htop-dev/htop@6fbee8c

Issue #878, opened 2019-01-11
make "htop -u" the shortcut for "htop -u $USER"


guettli:
It would be very nice, if this command:

    htop -u $USER

would be the same as this one:

    htop -u

This would be simpler.

Thank you.

honggyukim:
If it's not good to reuse existing `-u` option because it always have to accept an argument, the following can be a good option as well.
```
htop -U
```
Thanks.

guettli:
Yes, `htop -U` would be nice.

PR #884, opened 2019-01-30
User option "-u" now defaults to $USER

solanav:
The changes make the "-u" option have an optional argument that defaults to the user running htop. It was proposed in #878.


natoscott:
Merged here: htop-dev/htop@fb679df

PR #886, opened 2019-02-07
Honour setting of counting CPUs from 0/1 when presenting meters

rnsanchez:
In the listing of Available Meters for CPUs, the list of CPUs is always
presented by counting them from one.  However, if the user prefers to
count CPUs from zero, this is sometimes confusing when fine-tuning the
meters.


natoscott:
Merged here: htop-dev/htop@6310e5d

PR #890, opened 2019-02-13
Add pressure stall information (PSI) meters on Linux

### LINKED ISSUE #879: Handle /proc/pressure
Linux 4.20 adds `/proc/pressure/`, which contains some information about the pressure on CPU, memory, and i/o.

You can read about this here, among others: https://lwn.net/Articles/759781/

That would be great to have those metrics handled by htop, just like the load average.

bluetech:
The pressure stall information (PSI) metrics provide useful information
on delays caused by waiting for CPU, IO and memory. Particularly on busy
servers it can provide a quick overview of what's "slowing things down".

This feature is supported on Linux >= 4.20.
The interface is documented here:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/plain/Documentation/accounting/psi.txt
These links provide rationale:
https://lwn.net/Articles/759781/
https://facebookmicrosites.github.io/psi/

The following metrics are added, corresponding to the currently exposed
lines (see `head /proc/pressure/*`):
- PressureStallCPUSome
- PressureStallIOSome
- PressureStallIOFull
- PressureStallMemorySome
- PressureStallMemoryFull

The color scheme is the same as that used for Load Average, however I
gave it separate entries just in case someone wants to change them
specifically.

Tested on 4.20.7-arch1-1-ARCH, on the linux platform.
Also tested that other platforms still compile (changed configure to use
the unsupported platform).

Closes #879.


natoscott:
Merged here: htop-dev/htop@2be0992

PR #904, opened 2019-04-17
Truncate overwide jail names on BSD.

RobCrowston:
Jail names longer than eleven characters spill into adjacent columns and mess up the alignment.


natoscott:
Merged here: htop-dev/htop@e833bf9

PR #906, opened 2019-04-29
Add timestamps to the strace screen

mharjac:
Very useful for troubleshooting...


natoscott:
Merged here: htop-dev/htop@42c8e47

PR #914, opened 2019-05-27
Solaris 11 compatibility

senjan:
This PR aims to address 2 issues with htop on current Solaris:
1) htop does not build because Solaris's sys/sysinfo.h defines CPU_KERNEL which collides with definition within htop
2) when run in a Non-Global Zone, kstat module "zones" is not available thus kstat_lookup() returns NULL making htop to crash due to NULL pointer dereference in SolarisProcessList_readZoneName().


senjan:
fixed missing symbol when building on Linux

ninefathom:
Thanks, @senjan - good catch.  Keeping track of changes on both illumos and Oracle Solaris can be quite a chore.  Contributions are always appreciated!

dodona2:
is supporting an outdated OS really that important?

RobCrowston:
@dodona2 Solaris 11.4 is only nine months old. There are still pockets of industry users out there.

natoscott:
Merged here: htop-dev/htop@b30c22d (though I renamed the somewhat cryptic "CPU_KRNL" to "CPU_SYSTEM")

PR #917, opened 2019-06-23
Close pipe after lsof

Jesin:
When `htop` runs `lsof`, it leaves the read end of the pipe open. You can see this by selecting the htop process itself, pressing `l`, and then pressing `F5`/`Refresh` repeatedly. Every time it refreshes, there's 1 more pipe fd in the table until `RLIMIT_NOFILE` is reached. This commit fixes that.


Jesin:
Note that this commit also fixes a memory leak, since `fdopen` and `fclose` internally use `malloc` and `free` respectively (at least on some implementations).

natoscott:
Merged here: htop-dev/htop@74d5476

PR #153, opened 2014-12-23
Add ZFS cache awareness to memory meters.

Traviis:
Enable with --enable-zfs-cache-awareness

(Only tested with ZFS on linux)

PR #920, opened 2019-07-06
ZFS ARC stats for Linux, FreeBSD, Darwin, & Solaris

overhacked:
Added a new `Meter` type that displays ZFS ARC stats like FreeBSD's [top(1)](https://www.freebsd.org/cgi/man.cgi?query=top).
Supported on FreeBSD, Darwin/OS X/macOS, and Linux. FreeBSD & Darwin are nearly identical. Doesn't break or affect other OSes, as it is only included in `Platform_meterTypes` for supported OSes. There is also a potential buffer overflow bug fix in the first commit.


overhacked:
Hmm, I see that the build system tries to build it anyway for other platforms. I'll do some more work to make sure that it does not break the build for unsupported platforms.

ninefathom:
...I love it when my to-do list gets shorter on its own.  Thank you!

Mno-hime:
Works great on OpenIndiana 2019.04 in global and non-global zones. We will ship it via sources from @ninefathom soon. Thanks, @overhacked!

Mno-hime:
FYI: https://github.com/hishamhm/htop/issues/937.

overhacked:
Should be reasonably simple. I think visually it would need to be a different meter type. @ninefathom, will you still merge if I add another "ZFS ARC Compression" meter type?

ninefathom:
Sure would.

overhacked:
Resolving CI issues…

overhacked:
Tests out on FreeBSD, Linux, OmniOS, and Darwin.

ninefathom:
@Mno-hime release with the compressed ARC stats in the usual place, as 220-sunos_11-p3

overhacked:
Closes #937

despair86:
does this separate the ZFS ARC size from the rest of resident core set? As of now default settings allow the ARC to fill all unused core and this condition is impossible to tell apart in current htop: 
image

overhacked:
@despair86, I didn't mess with the code that calculates other types of memory usage in this PR, but I agree that it is more useful to have htop report ARC as cache usage. That [code already exists for the FreeBSD platform](https://github.com/hishamhm/htop/blob/666e1e76b39ee66a38f5fb620d22b23f36859eca/freebsd/FreeBSDProcessList.c#L324) but not for any other platforms.

I could add the tweaking of reported usage for Linux and Solaris/Illumos as a part of this PR. @ninefathom, any feedback? Would you rather see those changes in a separate PR, maybe an expansion of #976?

Harvie:
This is needed so badly. There are several PRs addressing this issue, We should decide which one is most useful:

#153 #920 #976 #937

natoscott:
Merged here: htop-dev/htop@9a55efc

PR #921, opened 2019-07-09
Fix for not completely updating command line

### LINKED ISSUE #905: Refresh process name doesn't work
Hisham,

It looks like `htop` doesn't always properly update process names when the corresponding option is enabled (as suggested in #672). Specifically I observe this happening with postgresql processes. To me it feels like it refreshes the name but the remainder of that line retains what it displayed previously.

Take a look at the screenshot I attached. 
htop-refresh-names

Notice that some `postgres` processes end with "`idle`" but others end with "`idle in transaction`" or "`idleCT`". Those that display "`idleCT`" previously displayed "`SELECT`" and after the name changes it refreshes the 1st 4 characters but retains "`CT`" after "`idle`". Same applies to those that end with "`idle in transaction`". When the name changes to end with "`idle`" the remainder "` in transaction`" isn't cleared.

Environment:
$ uname -a
Linux HOSTNAME 4.15.0-46-generic # 49-Ubuntu SMP Wed Feb 6 09:33:07 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

$ htop --version
htop 2.1.0 - (C) 2004-2018 Hisham Muhammad

Komzpa:
Expectedly fixes #905. Didn't check. strncpy reports itself as non-zero-terminated safe.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -670,12 +670,8 @@ static void LinuxProcessList_readDelayAcctData(LinuxProcessList* this, LinuxProc
 #endif
 
 static void setCommand(Process* process, const char* command, int len) {
-   if (process->comm && process->commLen >= len) {
-      strncpy(process->comm, command, len + 1);
-   } else {
-      free(process->comm);
-      process->comm = xStrdup(command);
-   }
+   free(process->comm);
+   process->comm = xStrdup(command);
    process->commLen = len;
 }
 


14mRh4X0r:
This is not the problem; all callers of `setCommand` ensure that `command[len] == '\0'`.

14mRh4X0r:
… except the one in `LinuxProcessList_readCmdlineFile`, bug introduced in bd5d37f297fed2a42acbbd730e2af5ad45af6afa

Issue #900, opened 2019-03-29
Ability to disable mouse


matecrn:
I generally prefer my mouse off in terminal programs, so it would be a nice addition. Most terminal programs have an on/off mouse function too, which is surprising that htop doesn't include this function. 

Unless there is already a workaround for this, I can't find any information on this topic.

PR #923, opened 2019-07-12
Added an option to disable the mouse.

MartinJM:
I saw Issue #900 and decided to take a look at it.

I have added an option in the Settings that stores if the mouse should be enabled or not. This can be toggled using a command line flag ("-m" or "--no-mouse"). If you want to permanently disable the mouse, the setting can also be set in the config file.

Apologies for the edits on the empty lines. The editor that I used removes whitespace if the line is empty otherwise. This is my first pull request, so if there is something I can do better, please let me know so I can learn from it.


MartinJM:
I found something I hadn't considered before. When starting the program with the flag to disable the mouse, and then changing another option within htop, the setting for the mouse is saved to the settings file as well. This means the mouse is disabled without an easy method to turn it on again.

For this reason, I have added the option to the display options as well.

natoscott:
Merged here: htop-dev/htop@011125d

Issue #671, opened 2017-08-21
Cross-compilation using Yocto fails on ncurses check


cmcqueen:
I'm trying to use Yocto to cross-compile htop 2.0.2 for an ARM-based system (BeagleBone Black). I'm using configure option `--disable-unicode` since I have disabled wchar support.

It fails with the message:

> ...
> checking for refresh in -lncurses... no
> checking for refresh in -lncurses6... no
> checking for refresh in -lncurses... (cached) no
> checking for curses.h... (cached) yes
> configure: error: missing libraries:  libncurses

The `config.log` contains:

> configure:13533: checking for refresh in -lncurses
> configure:13558: arm-poky-linux-gnueabi-gcc  -march=armv7-a -marm -mfpu=neon  -mfloat-abi=hard --sysroot=/home/craigm/yocto-laura/poky/build/tmp/sysroots/laura -o conftest  -O2 -pipe -g -feliminate-unused-debug-types -fdebug-prefix-map=/home/craigm/yocto-laura/poky/build/tmp/work/armv7ahf-neon-poky-linux-gnueabi/htop/2.0.2-r0=/usr/src/debug/htop/2.0.2-r0 -fdebug-prefix-map=/home/craigm/yocto-laura/poky/build/tmp/sysroots/x86_64-linux= -fdebug-prefix-map=/home/craigm/yocto-laura/poky/build/tmp/sysroots/laura=   --should-not-have-used-/usr/bin/ncurses5-config -Wl,-O1 -Wl,--hash-style=gnu -Wl,--as-needed conftest.c -lncurses  -lm  >&5
> arm-poky-linux-gnueabi-gcc: error: unrecognized command line option '--should-not-have-used-/usr/bin/ncurses5-config'

See [[oe] --should-not-have-used-](http://lists.openembedded.org/pipermail/openembedded-devel/2015-February/100066.html) on the OpenEmbedded mailing list which describes the `should-not-have-used-` part in the error message.

htop 1.0.3 builds okay in Yocto, I guess because it's not using the same method to check for ncurses.

wjx0912:
htop 2.2.0, the same problem, not fixed yet?

marv:
We hit the same issue on Exherbo. For now I was able to workaround that by passing
```
HTOP_NCURSESW_CONFIG_SCRIPT="/usr/$(exhost --target)/bin/ncursesw6-config"
```
[...]

PR #931, opened 2019-08-04
configure.ac: Use pkg-config to check for ncurses

Polynomial-C:
in order to correctly link against ncurses' tinfo lib if available
separately.

Gentoo-bug: https://bugs.gentoo.org/690840
Signed-off-by: Lars Wendler <polynomial-c@gentoo.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -22,6 +22,18 @@ AM_INIT_AUTOMAKE([1.11])
 AC_PROG_CC
 AM_PROG_CC_C_O
 
+m4_ifdef([PKG_PROG_PKG_CONFIG], [
+   PKG_PROG_PKG_CONFIG()
+], [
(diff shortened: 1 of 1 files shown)


marv:
I can confirm that this works as expected on Exherbo and fixes cross-compiling (https://github.com/hishamhm/htop/issues/671)

marv:
Can anyone shed some light on the situation on MacOS? I guess that's what's keeping this from getting accepted?

natoscott:
@Polynomial-C @marv there is no pkg-config by default on Mac OS X.  I see @hishamhm has created a  "pkgconfig-on-linux" branch - could you review and see if this meets your needs here?  Thanks.

[ see also https://github.com/hishamhm/htop/issues/992 as to why I'm contacting you - cheers. ]

PR #946, opened 2019-09-07
Linux: fixes sysfs battery discovery

smattie:
Removes the assumption that batteries are found in `/sys/class/power_supply/BAT?/` and instead reads from `/sys/class/power_supply/*/type`


natoscott:
Merged here: htop-dev/htop@cdff8ae

PR #949, opened 2019-09-12
suggestion: make MainFunctions less densely packed

duchampdev:
Currently, for example 'search' and 'filter' look very tightly put together. This might confuse some users which F{1-10} relates to which function.
To resolve this while preserving equal spacing, increase all functions width to seven characters.


natoscott:
Merged here: htop-dev/htop@e25ac85

Issue #98, opened 2014-05-23
Vim-like movement


Vndtta:
Hi, I've searched in the archive of the mailing list and didn't find anything... Sorry if it's been asked before (I guess it must have)

Is there anyway of setting the movement commands like vim? hjkl?

Thank you

e9t:
One vote up +1 for this wish!

zkanda:
Hello, was this ever implemented?

Thanks

hishamhm:
it was added as Alt+{h,j,k,l}.

zkanda:
Hmm, weird, my version of htop is 2.0.1, but it doesn't seem to work.

kaefer:
It works (here on my machine with german keyboard layout) with pressing the left-alt button (alt and not alt-gr)

hishamhm:
@zkanda it might be a limitation of your terminal. Open a shell, type `read` and press Enter. It will give you a prompt that will echo the ANSI codes for your keypresses. Type `hjkl` with and without Alt. It _should_ produce different results. If the results are the same, htop has no way to detect that Alt was pressed. In my system it shows:

```bash
hisham@pointer ~]read
hjkl^[h^[j^[k^[l
```

zkanda:
My terminal seems to be working as expected. Using gnome-terminal.

Can you please confirm this behavior?
https://asciinema.org/a/6kvkdo6szprhqcrpi7z1ctn2e

hishamhm:
@zkanda That's odd! Keys from `read` are working as expected. Could you try it on another terminal you might have installed (`xterm`)? It might be an ncurses configuration issue? I'm at a loss here.

darvelo:
Would it be possible to set this as a config variable at compile-time or preferably on the command line? I use Alt+{h,j,k,l} for moving through windows in my tiled window manager.

zkanda:
Okay, I tried it on `xterm` without any configuration and it seems to worsen the situation. `read` is printing weird non ascii characters.
Here's the result: https://asciinema.org/a/2mmhmyenvhg97awgcxtn3hkjf

I'm on `ArchLinux` with `Gnome` as my DE if that matters.

mudox:
I use htop on OS X 10.11.5, I tested underf following combinations:
    - Terminal.app + [zsh / bash] + htop
    - iTerm.app + [zsh / bash] + htop
    - iTerm.app + tmux + [zsh / bash] + htop
All with no luck.
[...]

PR #959, opened 2019-10-31
Add simple vim mode

lytedev:
This PR adds a "vim_mode" setting (false/`0` by default) that causes keys to be remapped in the following way by the `ScreenManager`:

```
+ h -> LEFT
+ j -> DOWN
+ k -> UP
+ l -> RIGHT
+ LEFT  -> h (toggle help)
+ DOWN  -> j (noop)
+ UP    -> k (open kill menu)
+ RIGHT -> l (lsof current process)
+ K (Shift+K) -> k (open kill menu)
+ J (Shift+J) -> K (toggle show/hide kernel threads)
+ L (Shift+L) -> l (lsof current process)
```

I couldn't figure out where the manpage documentation is in the repo, though I admittedly did not look particularly hard.

I believe this change would be a welcome option for heavy vim users like myself who would like a familiar way to get around in htop.


AtomToast:
I would like to have an option for vim bindings in normal htop though I prefer the shortcuts implemented by this fork: https://aur.archlinux.org/packages/htop-vim-git  
Source code is here https://github.com/KoffeinFlummi/htop-vim

lytedev:
I'll have to check that out! My implementation has issues with the may I remap the keys so that typing `k` when searching puts the character `K` into the input string, so this implementation is not particularly good at all. :laughing:

andys8:
See https://github.com/hishamhm/htop/issues/98

Would be nice to have a well maintained and packaged fork with vim bindings.

nycex:
I like this patch overall, though I would leave the keys to what they do at the moment and do something like:
```
? -> h
q -> k
L -> l
```
instead, remapping the arrow keys like that wouldn't be very intuitive in my opinion.

andys8:
Btw there are other system monitors with vim bindings. I'm currently using `bottom` because it's configurable. There is also `ytop` and more. Depends on what you need.

https://github.com/ClementTsang/bottom
https://github.com/cjbassi/ytop

vcfvct:
> Btw there are other system monitors with vim bindings. I'm currently using `bottom` because it's configurable. There is also `ytop` and more. Depends on what you need.
> 
> https://github.com/ClementTsang/bottom
> https://github.com/cjbassi/ytop

Thanks for the suggestions. Can finally switch to a modern one with more natural key-bindings. `Bottom` looks great.

natoscott:
Merged here: htop-dev/htop@5c5a599

PR #960, opened 2019-10-31
Fix trailing whitespace

lytedev:


natoscott:
Merged here: htop-dev/htop@a82fd26

PR #964, opened 2019-11-10
Added missing option in manual

polluks:


natoscott:
Merged here: htop-dev/htop@1b8bed1

Issue #930, opened 2019-08-01
Zombies that don't have a parent-process are not shown


UlrichEckhardt:
Zombies without a parent process occur when running without an init process that reaps the orphaned child processes. Normally, this shouldn't occur, because orphaned children are adopted by init (PID 1) which reaps those orphans on termination. However, there are two scenarios where this can happen:

- Inside of a Docker container: It is not usual that containers are started without an init process. This is also the place where I stumbled across this problem.
- rescue mode: Instead of running init on boot, you boot into a shell in order to repair a broken system. Actually I haven't verified this part, but it's likely affected, too.

Note that I'm guessing what's causing the differences between the `ps auxf` output and that of `htop`, but I'd say it's a reasonable explanation of what I can observe.

joder:
I just encountered a similar problem on linux. However, in my case the zombie process is shown when showing kernel threads, but not otherwise. When digging into it with a debugger, I noticed that the `cmdline` file for the process is empty. However, if this file is empty then in the function `LinuxProcessList_readCmdlineFile` in line [695](https://github.com/hishamhm/htop/blob/402e46bb82964366746b86d77eb5afa69c279539/linux/LinuxProcessList.c#L695) the process is redefined as a kernel thread.

I hope this helps.

PR #970, opened 2019-12-14
Properly identify zombie processes

joder:
I suggest the following small fix with the following reasoning:

- the conditionals with `amtRead` in function `LinuxProcessList_readCmdlineFile` are meant to protect the execution of the for loop (and what follows it), which is not supposed to be run with `amtRead == 0` (if I understand correctly),
- the command for zombie processes is then set in function `LinuxProcessList_recurseProcTree` at line [883](https://github.com/hishamhm/htop/blob/402e46bb82964366746b86d77eb5afa69c279539/linux/LinuxProcessList.c#L883) after the "offending" call to `LinuxProcessList_readCmdlineFile` at line [856](https://github.com/hishamhm/htop/blob/402e46bb82964366746b86d77eb5afa69c279539/linux/LinuxProcessList.c#L856).

To me this is a quick fix and the setting of command name of processes seems a little bit clumsy (i.e. `readCmdlineFile` is not just reading, but actually setting the command name of a process and classifying it as a kernel thread), however, I lack the insight as well as experience to write it better. Maybe setting the command name of a zombie process should also be moved to `readCmdlineFile` and the function renamed. Please, let me know what you think about it.

This closes issue #930.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -692,7 +692,11 @@ static bool LinuxProcessList_readCmdlineFile(Process* process, const char* dirna
    int tokenEnd = 0; 
    int lastChar = 0;
    if (amtRead == 0) {
-      ((LinuxProcess*)process)->isKernelThread = true;
+      if (process->state == 'Z') {
+         process->basenameOffset = 0;
+      } else {
+         ((LinuxProcess*)process)->isKernelThread = true;
+      }
       return true;
    } else if (amtRead < 0) {
       return false;


joder:
I see now the fix for issue #930 should also probably be connected to resolution of issues #905, #921 and #812. Especially [this](https://github.com/hishamhm/htop/pull/921#issuecomment-515373548) comment by @14mRh4X0r made me think my proposed resolution is incorrect.

natoscott:
Merged here: htop-dev/htop@fed14a5

PR #971, opened 2019-12-17
update man page text about delay

davemq:
It has bothered me for quite some time that running `htop -d 590` does not delay 59 seconds but something much less. I finally looked at the code today and see that the delay is clamped to `1 <= delay <= 100', i.e. a minimum of 0.1 seconds and maximum of 10.0 seconds. It's not clear to me in looking at the code history if there's some design reason for this, so I propose this change to the manual page. I'd be happy to change the code to remove this restriction and revert this proposed change to the manual page, but I think that requires a bit more discussion.

### DIFF
--- a/htop.1.in
+++ b/htop.1.in
@@ -22,7 +22,9 @@ Mandatory arguments to long options are mandatory for short options too.
 .LP 
 .TP
 \fB\-d \-\-delay=DELAY\fR
-Delay between updates, in tenths of seconds
+Delay between updates, in tenths of seconds. If the delay value is
+less than 1 it is increased to 1, i.e. 1/10 second. If the delay value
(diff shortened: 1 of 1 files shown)


davemq:
Looking a bit further into this, the delay value eventually gets used in a call to `halfdelay()` in ncurses. The [halfdelay(3)](https://linux.die.net/man/3/halfdelay) man page says

> The halfdelay routine is used for half-delay mode, which is similar to cbreak mode in that characters typed by the user are immediately available to the program. However, after blocking for tenths tenths of seconds, ERR is returned if nothing has been typed. The value of tenths must be a number between 1 and 255. Use nocbreak to leave half-delay mode.

natoscott:
Merged here: htop-dev/htop@f4b0067

PR #974, opened 2019-12-29
Fix STARTTIME column on FreeBSD.

RobCrowston:
Before this patch, the "START" column is a zero-width string. This also causes subsequent columns to be misaligned with their titles.


RobCrowston:
Resolved a bug where the start time is not amended after it is >24 hours old (amended commit).

natoscott:
Merged here: htop-dev/htop@ab61ae3

PR #976, opened 2020-01-11
Add ZFS cache awareness to memory meters.

### LINKED ISSUE #44: ZFS buffer/cache is counted under used memory
I'm running zfsonlinux and htop is counting the ZFS ARC (memory) cache under normal used memory.

This is because ZFS doesn't use normal Linux memory cache but it's own version.

It would be useful if htop could detect if ZFS is running (maybe checking for /dev/zfs) and in that case show the memory used by ZFS cache in a different colour or maybe joined with the normal buffers/cache.

ARC used memory can be checked in /proc/spl/kstat/zfs/arcstats (size):

```
# grep ^size /proc/spl/kstat/zfs/arcstats 
size                            4    3533162592
```

zielmicha:
Fixes #44

This is #153, but rebased to be mergable with the current head.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -233,6 +233,11 @@ else
             [AC_CHECK_HEADERS([ncurses.h],[:],[missing_headers="$missing_headers $ac_header"])])])])
 fi
 
+AC_ARG_ENABLE(zfs_cache_awareness, [AC_HELP_STRING([--enable-zfs-cache-awareness],[enable ZFS cache awareness for memory usage])], enable_zfs_cache_awareness="yes",enable_zfs_cache_awareness="no")
+if test "x$enable_zfs_cache_awareness" = xyes; then
+    AC_DEFINE(HAVE_ZFS, 1, [Define if we want the memory meter to be aware of the ZFS cache])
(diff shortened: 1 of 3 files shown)


natoscott:
@zielmicha  I believe this PR will need to be reworked to fit in with the code from https://github.com/hishamhm/htop/pull/920 

It should simplify the changes needed here immensely - the arcstats file parsing is done and there's no need for configure.ac changes, so *most* of this PR goes away (put not all AFAICT).

We're creating a community 3.0.0 htop release, it'd be great to have this functionality included for Linux ZFS.
[ https://github.com/hishamhm/htop/issues/992 ]

I notice the FreeBSD code (freebsd/FreeBSDProcessList.c) already handles the memory accounting adjustments you're making here - but from my understanding of the code in both PRs, I believe the functionality you seek here will still be missing for Linux.

Please see https://github.com/htop-dev/htop for a tree with #920 merged - if someone with ZFS filesystems could create a new PR ASAP for the htop-dev repo, it'd be great to include this in that new release.  Thanks!

PR #981, opened 2020-01-23
Fix gcc10

SoapGentoo:
Hi @hishamhm 
GCC 10 switches from `-fcommon` to `-fno-common` by default (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85678), and htop currently fails when building against `-fno-common`.

I've enabled `-fno-common` in Travis to emulate GCC 10's behaviour, to catch any future bugs like this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,4 @@ os:
   - linux
   - osx
 
-script: ./autogen.sh && ./configure && make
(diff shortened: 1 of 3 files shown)


ginggs:
See line 1 of **CRT.h** and **linux/LinuxProcess.h** :

`/* Do not edit this file. It was automatically generated. */`

PR #982, opened 2020-01-25
htop: fix build failure against gcc-10

trofi:
On gcc-10 (and gcc-9 -fno-common) build fails as:

```
gcc -pedantic ... -o htop ...
ld: ClockMeter.o:/home/slyfox/dev/git/htop/CRT.h:159:
  multiple definition of `backtraceArray'; CheckItem.o:htop/CRT.h:159: first defined here
ld: ClockMeter.o:/home/slyfox/dev/git/htop/CRT.h:153:
  multiple definition of `CRT_termType'; CheckItem.o:htop/CRT.h:153: first defined here
```

gcc-10 will change the default from -fcommon to fno-common:
https://gcc.gnu.org/PR85678.

The error also happens if CFLAGS=-fno-common passed explicitly.

Reported-by: Toralf Förster
Bug: https://bugs.gentoo.org/706358

### DIFF
--- a/CRT.h
+++ b/CRT.h
@@ -140,7 +140,7 @@ extern const char **CRT_treeStr;
 
 extern int CRT_delay;
 
-int* CRT_colors;
+extern int* CRT_colors;
 
 extern int CRT_colorSchemes[LAST_COLORSCHEME][LAST_COLORELEMENT];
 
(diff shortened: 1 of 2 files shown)


SuperSandro2000:
Dupe of #981

PR #985, opened 2020-02-03
Introduce 4-column CPU Meter option

zlim:
Enable this option on Linux, when CPU count > 32 by default.


zlim:
@hishamhm would you consider pulling this please? Thanks :)

natoscott:
@zlim https://github.com/hishamhm/htop/pull/1012 seems to be a slightly more general solution to the same problem you've tackled here.

PR #996, opened 2020-04-27
Fix "Broken Gray" didn't change the color of cpu-iowait

fo40225:
Fix "Broken Gray" didn't change the color of cpu-iowait.


natoscott:
Merged here: htop-dev/htop@54b6143

PR #1004, opened 2020-05-27
fix build with GCC 10.x

eworm-de:
Building with GCC 10.x fails with lots of:

multiple definition of ...

Defining these extern fixes the build.


eworm-de:
Well... Another duplicate of #981...
Any chance to get some pull requests merged?

PR #1011, opened 2020-06-11
Let the user know about their error

jpereira:
If the user informed wrong value, then let them know about that. At least, print out the error and leave the program.


natoscott:
Merged here: htop-dev/htop@cca07e9b

PR #1012, opened 2020-06-11
adding support for more than 2 smaller cpumeter columns

cbudziszewski:
Hello,

since CPU count goes up and monitors tend to get wider rather than taller, more cpumeter columns free up more space for the process list.
This PR adds two functions, that will allow an arbitrary number of columns as well as classes for displaying 4 smaller columns.

Kind regards,
Christoph Budziszewski


natoscott:
Merged here: htop-dev/htop@31391b2

PR #1024, opened 2020-08-20
Fix the error that the period variable obtained for the first time is abnormal

Nacho-meteor:
Hello, when I was watching and learning the source code, I found that the period variable data obtained for the first time was abnormal, so I judged the variable data of this-> cpus[0].totalPeriod,Fix the error that the period variable obtained for the first time is abnormal.

### DIFF
--- a/linux/LinuxProcessList.c
+++ b/linux/LinuxProcessList.c
@@ -927,30 +927,30 @@ static inline void LinuxProcessList_scanMemoryInfo(ProcessList* this) {
    char buffer[128];
    while (fgets(buffer, 128, file)) {
 
-      #define tryRead(label, variable) (String_startsWith(buffer, label) && sscanf(buffer + strlen(label), " %32llu kB", variable))
+      #define tryRead(label, variable) do { if (String_startsWith(buffer, label) && sscanf(buffer + strlen(label), " %32llu kB", variable)) { break; } } while(0)
       switch (buffer[0]) {
       case 'M':
-         if (tryRead("MemTotal:", &this->totalMem)) {}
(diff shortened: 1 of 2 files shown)


natoscott:
This change looks like a workaround rather than a root cause fix?  I guess the question is, *why* the 'period variable obtained for the first time is abnormal'... any idea?

I wonder if the rapid-fire double sampling done in htop.c main() is part of the story here:

```
   ProcessList_scan(pl);
   millisleep(75);
   ProcessList_scan(pl);
```

could we end up with an abnormal period value due to this perhaps?  (we need this code so that counters can be rate-converted)  It is using a (short) internal (75msec) different to the regular sample interval - maybe a calculation somewhere uses the wrong time interval?  sorry, bit of a wild guess :) - but we do need to understand the underlying problem properly still I think.

Nacho-meteor:
Because when I was using htop, I found that the CPU utilization in the first second after opening the program was displayed as 100%. When I tried to solve this problem, I found that the output variable period was abnormal, and only the first abnormally large , So I first wanted to rule out this situation, so I made this change.

Nacho-meteor:
I thought about your guess carefully, I think it makes sense, I will confirm it in the direction of your guess, thank you for your reply

natoscott:
> I thought about your guess carefully, I think it makes sense, I will confirm it in the direction of your guess, thank you for your reply

@Nacho-meteor no problem at all.

There definitely seems to be an underlying issue here somewhere - I've seen quite a few people mention it now; keep at it, I'm sure you'll figure it out.

